PR #49, opened 2017-11-22
User profile

### LINKED ISSUE #25: Create user profile tables
The base Django user model only has username, first and last name, and email address. The Django user model is easily extended to include additional information (called a User Profile). 

The existing database has extra user info, such as additional contact info among other things. If we need to create user profiles we need to know which fields of the original database should be carried over.

dereklieu:
Close #25 

- Creates a user profile one-to-one relation with users
- Ingests user data from the URLs.mdb user table
- Doesn't ingest users who haven't logged in for a year
- Creates routes to serve user profiles. `profile?user={pk}` returns both user profile and user table data.


dereklieu:
@matthewhanson this is now running and I've confirmed the ingest works. Would still appreciate it if you took a quick glance through though.

PR #57, opened 2017-11-28
Index to elasticsearch

### LINKED ISSUE #51: Index documents to ES
To start, index page titles for each data page (field report, appeal, event, etc) along with the type of page.

dereklieu:
Close #51 

- Creates an elasticsearch client.
- Connect indexing function to `post_save` so indexes update automatically.
- Disable automatic indexing on bulk imports and adds an `index_elasticsearch` command to use instead.
- Adds serializers to relevant models to create elasticsearch documents.

@matthewhanson I'm not crazy about using the env variable to disable automatic indexing during bulk uploads, but also couldn't think of a better way.

It's also possible that the fields I've chosen to index don't prove super useful, but given the size of the data, it's pretty quick to delete and re-index the entire dataset. Right now I'm not sure if putting everything into the same index is the right idea, for example.


dereklieu:
@matthewhanson failing tests as there's no local ES running and I don't want to connect to the production instance. What do you think about running the tests in a docker environment where elasticsearch is present, and we can just kick it off locally?

matthewhanson:
@dereklieu We can do that, but we could also mock the ES connection in the tests.  It might be good to keep unit tests simple and separate from integration tests (where we could test things like ES)

dereklieu:
@matthewhanson pushed a change that requires an `ES_HOST` env var before it tries to connect. That felt like the least invasive way of doing it.

PR #58, opened 2017-11-28
FieldReport and Appeal updates

matthewhanson:


matthewhanson:
work in progress, need to finish up contacts and fix failing tests

api/management/commands/ingest_appeals.py
```diff
@@ -63,10 +63,12 @@ def handle(self, *args, **options):
                 country = country.first()
 
             appeals = [a for a in r['Details'] if a['APD_code'] not in aids]
+            atypes = {66: AppealType.DREF, 64: AppealType.APPEAL}
```
dereklieu: Noting that the last type is 'international and it will be code `1537`. I will add this in.

PR #71, opened 2017-11-29
Turn off debug for json error messages

dereklieu:
@danielfdsilva I tried a few things to wrap the default error response but it was more trouble than I anticipated. However, turning off `debug` seems to get us close to what we want. Instead of the original messages, in a few tests I've gotten this instead:

`{"error_message": "Sorry, this request could not be processed. Please try again later."}`

I can switch over the other messages to mimic this but don't think I can expend more effort trying to change these now.


danielfdsilva:
@dereklieu This does the trick. It's JSON, we can then show the appropriate message based on the status code and context

dereklieu:
@danielfdsilva K, merging this. I've switched the other endpoints (ie the auth token endpoint) to use `error_message` instead of `message` now as well.

Issue #77, opened 2017-12-01
Connect API to email notification system


dereklieu:
Creations and edits of relevant objects should trigger email notifications to user who have subscribed to these notifications.

This will entail creating a separate app for notifications, and a model to hold these subscription preferences.

PR #79, opened 2017-12-01
WIP Connect email notifications

dereklieu:
Address #77 

Todo's:

- [x] Add regions to country fixture data
- [ ] Hook up to `POST_SAVE` hook for relevant subscriptions
- [ ] Create views to populate emails for the various types of subscriptions
- [ ] Sync the DMIS subscriptions table so that people don't lose their subscriptions (but create a flag so we don't send tons of emails yet while DMIS is still active)
- [x] Add an API route to toggle subscriptions on and off
- [ ] Add views to authenticate new users


dereklieu:
Syncing this now and ticketing out other items separately.

PR #84, opened 2017-12-03
Include related fields

dereklieu:
`/api/v1/field_report/` includes the full `event` (if it's attached)
`/api/v1/appeal/` includes the full `event`
`/api/v1/event/` includes full lists of `appeals` and `field_reports` (if attached)

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -3,7 +3,7 @@ from django.db import models
 from django.conf import settings
 from django.db.models.signals import post_save
 from enumfields import EnumIntegerField
-from enumfields import Enum
+from enumfields import IntEnum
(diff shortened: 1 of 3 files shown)


dereklieu:
@danielfdsilva This pr implements related fields in the important api responses (appeal, events, and field reports) and also includes a range of filtering options. For those, check `api/resources.py`, ie [here](https://github.com/IFRCGo/go-api/pull/84/files#diff-d7b1ed3f3558dd742a98c1435c3b5daaR102).

Example usage: `/api/v1/event/?created_at__gte=2017-11-3&limit=100`

PR #152, opened 2017-12-27
Sync allowed hosts

dereklieu:


main/settings.py
```diff
@@ -1,15 +1,20 @@
...
 localhost = 'localhost'
-BASE_URL = localhost if local_test else production_url
+
+BASE_URL = '%s:8000' % localhost if local_test else production_url
+
+ALLOWED_HOSTS = []
+if local_test:
+    ALLOWED_HOSTS.append(localhost)
+elif production_url is not None:
+    ALLOWED_HOSTS.append(production_url)
```
dereklieu: cc @matthewhanson this allows the `allowed_host` to get added to Django settings from an env file without hard-coding, to enable multiple deployments. This means two things:

1. When developing locally you'll have to set the `LOCAL_TEST=1` env var.
1. You won't be able to connect to the deployed DB from localhost anymore, because of this change, and the fact that the db is now firewall-restricted.

PR #186, opened 2018-01-17
add alert level to events + migrations, resolves #168

### LINKED ISSUE #168: Add alert level (red, orange, green) to events
Should also ingest this from the gdacs feed where available.

matthewhanson:


dereklieu:
@matthewhanson can we also make this an enum field? The only options should be red, orange, and green.

matthewhanson:
@dereklieu yes, np. Will update PR later today

matthewhanson:
@dereklieu turned into enum, synced up with develop and redid migrations.

however, it's failing tests on Circle as it's not recognizing the alert_level column.  I tried without the Circle cache and same problem so I"m not sure why it's complaining. Tests run fine locally.

I'm on my way out, but will revisit this again later.

matthewhanson:
@dereklieu this is ready for review and merge

PR #188, opened 2018-01-17
include inline appeal form under emergency (event), resolves #169

### LINKED ISSUE #169: Include DREF/Appeal admin form UI for emergencies

matthewhanson:


dereklieu:
@matthewhanson is there a form element that would allow an operator to link an appeal that already exists? I don't think a common use case is to create new appeals through the event object admin form. Appeals don't really get created in the admin at all in fact.

matthewhanson:
@dereklieu Not sure, I'll look into this.

matthewhanson:
@dereklieu From the Appeal form in the admin you can select an Event, but because Appeal has the ForeignKey to the Event you cannot add a select field. Event would have to have a ManyToMany relation with Appeal, but this would allow Appeals to be linked to multiple Events. You would see all the Appeals in the select box under Event.

This may be possible with some custom admin form elements, but is selecting the Event from the Appeal an acceptable work-around for now?

dereklieu:
Yeah I think that has to be the work-around. Thanks for looking into this.

PR #240, opened 2018-06-05
updates to how to build/develop and update the code base

scisco:
changes:
- new changelog
- updated docker-compose workflow
- updated circleci workflow
- getting rid of the develop branch
- update how the project is released


dereklieu:
Works great. Noting I had to remove an existing Postgres image before running migrate and getting this to work. 

Also I'm glad we're able to use the Postgres instance to test. That was one of the major bonus points for doing this in the first place. I've found that Sqlite doesn't throw on some of the things that Postgres will, and that has bitten us when we've deployed this in the past.

PR #272, opened 2018-07-04
Sitreptype

szabozoltan69:
Addresses @dereklieu 

## Changes

* Situation report type can be filtered by (doc) type
* Test script (test_views.py) written and works
* Admin registration of DocumentType table is done

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


dereklieu:
@szabozoltan69 you need to add and commit the new migration file as well. Otherwise the test will fail, since it is attempting to test with an old version of the database.

szabozoltan69:
Thank you for the guidance, I uploaded them (api/migrations/0008_auto_20180703_1559.py, api/migrations/0009_auto_20180704_0716.py) as well.

api/models.py
```diff
@@ -270,6 +275,7 @@ class SituationReport(models.Model):
     document_url = models.URLField(blank=True)
 
     event = models.ForeignKey(Event, on_delete=models.CASCADE)
+    type = models.ForeignKey(DocumentType, related_name='situation_reports', null=True, on_delete=models.CASCADE)
```
dereklieu: @szabozoltan69 since we're allowing `null=True`, probably easier for the user to just have `on_delete=models.SET_NULL` so that removing a document type doesn't wipe out all it's associated documents.

szabozoltan69:
Thank you for the lots of corrections and enhancements.

PR #293, opened 2018-09-18
Removing the required property from the region relation in the countr…

szabozoltan69:
## Changes

* Country region: inserting a "blank=True"

Solves #415


dereklieu:
@szabozoltan69 thanks, do you need to commit a migration with this in order for it to take effect?

szabozoltan69:
No, as I tested (without migration) it works fine now, the region can be
set to none (----).

2018. szept. 18. du. 6:36 ezt írta ("Derek Lieu" <notifications@github.com
>):

@szabozoltan69 <https://github.com/szabozoltan69> thanks, do you need to
commit a migration with this in order for it to take effect?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/IFRCGo/go-api/pull/293#issuecomment-422462397>, or mute
the thread

PR #306, opened 2018-10-04
Allowing different lower/uppercase lettered usernames in views.py

szabozoltan69:
Used https://stackoverflow.com/questions/13190758/django-case-insensitive-matching-of-username-from-auth-user

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -291,7 +291,18 @@ class GetAuthToken(PublicJsonPostView):
         username = body['username']
         password = body['password']
 
-        user = authenticate(username=username, password=password)
+        #allowing different lower/uppercase lettered usernames:
+        caseSensitiveUsername = username
+        try:
+            findUser = User._default_manager.get(username__iexact=username)
+        except User.DoesNotExist:
+            findUser = None
+        if findUser is not None:
+            caseSensitiveUsername = findUser
+
+        #user = authenticate(username=username, password=password)
+        user = authenticate(username=caseSensitiveUsername, password=password)
+
         if user is not None:
             api_key, created = Token.objects.get_or_create(user=user)
 


szabozoltan69:
Solves github issue #471 (https://github.com/IFRCGo/go-frontend/issues/471)

api/views.py
```diff
@@ -291,7 +291,18 @@ def handle_post(self, request, *args, **kwargs):
         username = body['username']
         password = body['password']
 
-        user = authenticate(username=username, password=password)
+        #allowing different lower/uppercase lettered usernames:
+        caseSensitiveUsername = username
+        try:
+            findUser = User._default_manager.get(username__iexact=username)
```
dereklieu: @szabozoltan69 is there a reason why you don't use `User.objects.get()` here? That is the standard way to do this query.

api/views.py
```diff
@@ -291,7 +291,18 @@ def handle_post(self, request, *args, **kwargs):
         username = body['username']
         password = body['password']
 
-        user = authenticate(username=username, password=password)
+        #allowing different lower/uppercase lettered usernames:
+        caseSensitiveUsername = username
```
dereklieu: @szabozoltan69 since the rest of the codebase uses `lower_case` variable and function naming, please follow that.

https://stackoverflow.com/questions/159720/what-is-the-naming-convention-in-python-for-variable-and-function-names

api/views.py
```diff
@@ -291,7 +291,18 @@ def handle_post(self, request, *args, **kwargs):
...
+        try:
+            findUser = User._default_manager.get(username__iexact=username)
+        except User.DoesNotExist:
+            findUser = None
+        if findUser is not None:
+            caseSensitiveUsername = findUser
+
+        #user = authenticate(username=username, password=password)
```
dereklieu: @szabozoltan69 please remove this unless you have a reason to keep it here. If you're worried about reverting this change, that should be handled by proper git history.

api/views.py
```diff
@@ -291,7 +291,18 @@ def handle_post(self, request, *args, **kwargs):
...
 
-        user = authenticate(username=username, password=password)
+        #allowing different lower/uppercase lettered usernames:
+        caseSensitiveUsername = username
+        try:
+            findUser = User._default_manager.get(username__iexact=username)
+        except User.DoesNotExist:
+            findUser = None
+        if findUser is not None:
+            caseSensitiveUsername = findUser
```
dereklieu: It's more straightforward here to just say 

`try: case_sensitive_username = User.objects.get(username__iexact=username)`
`except User.DoesNotExist: case_sensitive_username = None`

szabozoltan69: It would be the safest if you could do the change please. I just used a stackoverflow-found solution (this is the reason of the different variable format also).

szabozoltan69: This is already in master branch.

api/views.py
szabozoltan69: It seemed to be reasonable to keep here the "clean" (case sensitive) solution also. But it can be deleted also if it results nicer screen.

dereklieu: I don't agree. This code is outdated. It has been replaced by your solution. There is no reason to keep it when it already exists in git history.

api/views.py
dereklieu: You could just open another PR?

szabozoltan69: I created https://github.com/IFRCGo/go-api/pull/308 for this.

PR #316, opened 2018-10-11
We use all records, do NOT check if last_modified > since_last_checked

szabozoltan69:
A useful feature now omitted due to Natig's new financial API solution,
removing:
                since_last_checked = datetime.utcnow().replace(tzinfo=timezone.utc) - timedelta(minutes=90)
                last_modified = self.parse_date(r['APP_modifyTime'])
                if last_modified > since_last_checked:

### DIFF
--- a/api/management/commands/ingest_appeals.py
+++ b/api/management/commands/ingest_appeals.py
@@ -42,14 +42,12 @@ class Command(BaseCommand):
             with open('appeals.json', 'w') as outfile:
                 json.dump(records, outfile)
 
-            since_last_checked = datetime.utcnow().replace(tzinfo=timezone.utc) - timedelta(minutes=90)
             codes = [a.code for a in Appeal.objects.all()]
             for r in records:
                 if not r['APP_code'] in codes:
                     new.append(r)
-                last_modified = self.parse_date(r['APP_modifyTime'])
-                if last_modified > since_last_checked:
-                    modified.append(r)
+                # We use all records, do NOT check if last_modified > since_last_checked
+                modified.append(r)
 
         return new, modified
 
--- a/api/management/commands/ingest_appeals.py
+++ b/api/management/commands/ingest_appeals.py
@@ -42,14 +42,12 @@ class Command(BaseCommand):
             with open('appeals.json', 'w') as outfile:
                 json.dump(records, outfile)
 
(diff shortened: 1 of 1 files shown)


dereklieu:
@szabozoltan69 what do you think about specifying a time range say of the last 3 years, instead of all? Might be a good common ground. @mmusori thoughts on a time range where you no longer expect appeals to be updated?

szabozoltan69:
@dereklieu I can immagine this 3 years limit, yes.

szabozoltan69:
@dereklieu please approve this PR. The log of the running gives this:

mmusori:
Dear Derek
I would say four years in accordance with the extension limits in the EPoA procedures. The outliers are Haiti Earthquake which ran for 4,5 years Syria is ongoing 6,5 years later. And any new funding would be allocated to that Appeal. But let me consult with Natig if taking  information for only the last 4 or 6 years would affect transactional data or the child tables not linked to the master. 
Regards
Munu
    On Thursday, October 11, 2018, 5:06:47 PM GMT+2, Derek Lieu <notifications@github.com> wrote:  
 
 
@szabozoltan69 what do you think about specifying a time range say of the last 3 years, instead of all? Might be a good common ground. @mmusori thoughts on a time range where you no longer expect appeals to be updated?

szabozoltan69:
Ok, I set 4 years...

szabozoltan69:
If I change the limit from 3 years to 4 years, that is the difference between the logs:
```
Starting app... Starting appeals ingest
Querying app... Querying appeals API for new appeals data
3127 current... 3127 current appeals
Creating 0 n... Creating 0 new appeals
Updating 468... Updating 556 existing appeals that have been modified
0 appeals cr... 0 appeals created
468 appeals ... 556 appeals updated
3127 total a... 3127 total appeals
Appeals inge... Appeals ingest completed
```

mmusori:
Hi Derek

How much lag time would be caused by uploading the entire content of the API.
As I said earlier 4 to 6 years would be a good timeframe to capture all transactions but I still would like to confirm with Natig on what effects this would have.

Regards

Munu

szabozoltan69:
(There is no much run time difference between 3, 4 or 6 years limit. You should decide, @mmusori and I put it to the code.)

PR #319, opened 2018-10-15
Show emergency checkbox in backend admin

szabozoltan69:
For events we should allow or disallow emergency (right side) showing, e.g. for /emergencies/3364.


dereklieu:
@szabozoltan69 thank you. I think the field name could be more clear. For one, this is not hiding the emergency, it is hiding the field report. A suggestion:

`hide_attached_field_reports` with default to `False`.

Let me know if you have any trouble re-creating a new migration.

PR #329, opened 2018-11-12
Update Countries.json

ElsaRaunioIFRC:
Addresses xxxxxx 

## Changes

* ...

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


szabozoltan69:
What is needed for the successful circleci test is to replace the "count of countries" in api/test_models.py - line 50. The test asserts if we have 260 countries, it should be only 258.

Issue #437, opened 2019-06-24
Map Export v1


batpad:
This is a bit of a higher level ticket for the `v1` of the "Map Export" tool - we can break these out into smaller tickets if required.

For this iteration, we want to add functionality for the user to see a map on the Emergency Detail page, highlighting the affected country and `admin1` boundaries within it. They should also be able to "Export" this map - i.e. be able to export to an image for print, etc.

For this we will need:

 - Ability to associate Emergencies with `District`s in the database (Many2Many relation)
 - Ability for user to associate Emergencies with Districts in the form to add Field Reports.
 - Highlight the associated districts on the emergency pages, on a custom designed Mapbox map.
 - Repurpose the "Print Map" functionality that @mankamolnar has been working on to export the map and surrounding HTML / design elements.

We also probably want the ability to have logos of the National Society involved on the map. I *think* this would involve adding a `logo` field or so to the `Country` model: https://github.com/IFRCGo/go-api/blob/master/api/models.py#L71 - for now, we can probably add the logos in the Django Admin, so we should not need any frontend to enable National Societies to add logos.

cc @szabozoltan69 @ElsaRaunioIFRC

batpad:
I realize I misunderstood the database structure a little bit when writing the above ticket.

The current structure has `Emergencies` (this is `Events` in the database), and within it, it can have many Field Reports.

I imagine we want the map to be exportable for the `Emergency`, hence it might not make sense to have the step to choose the Districts as part of the Field Report form. Is there a separate step where `Emergencies` are created? Ideally, the districts would be specified to be associated with the Emergencies? And then we can show them on the map on the Emergency page as per the designs and make it exportable.

Let me know if this makes sense ^ and if this is something @szabozoltan69  either you or @ElsaRaunioIFRC can help with - i.e. figuring out where we store the association to Districts and also where will we ask users to add this data about which Districts this Emergency pertains to.

szabozoltan69:
Well, in the process of the Field Report creation there is a decision: do we link it to an existing Emergency (=Event) or not (in this case the backend will create a new emergency with the Field Report's name).

You are correct, the districts should be associated with Emergencies.

batpad:
@szabozoltan69 interesting.

So perhaps we need to give the user the option to add districts in the Field Report form, and have some logic like:

 - If `Emergency` (=Event) does not exist, create it and associate districts that user has added.
 - If `Emergency` exists, add any new districts user has selected in the Field Report form to it.

[...]

PR #457, opened 2019-07-03
add fields to Country and District

batpad:
This adds the `record_type` and `is_enclave` fields as per https://github.com/IFRCGo/go-api/issues/437#issuecomment-507611037

This also make the `record_type` field for Countries editable in the List View in the Admin, to make it easier for users to edit.

@szabozoltan69 let me know if you are able to test this and if this looks good. Please wait a bit for us both to test before merging - I just want to make sure adding these fields does not break anything else (I don't think it should, but will be good to be sure!)

@geohacker - can I also ask you for a quick review?


api/models.py
```diff
@@ -71,10 +71,21 @@ def region_name(self):
...
     return 'logos/%s/%s' % (instance.iso, filename)
 
+class CountryType(IntEnum):
+    '''
+        We use the Country model for some things that are not "Countries".
+        For eg. IFRC (Not a country) or groups of countries like Southeast Asia (A region, not a country)
+    '''
+    NOT_COUNTRY = 0
+    COUNTRY = 1
+    GROUPING = 2
```
geohacker: Is `grouping` the best way? Or should we store actual admin level type?

batpad: Yea, was not sure what to call this case -- so, we have entries that are Countries - eg. India. This is straightforward as type `1` above.

Groupings are things like:
 - `Andean Country Cluster`
 - `Asia Pacific Region`
 - `Central Asia Country Cluster`
 - `English Speaking Caribbean Country Cluster`

I'm not sure these fit neatly into admin level definitions either. Perhaps we can go with `Cluster` to be closer to how a lot of these are called in the backend?

@ElsaRaunioIFRC - do you have an idea for what "type" we should assign to things like the examples above?

ElsaRaunioIFRC:
Hi Sanjay,

How about this type of approach?

“ENTITY”, “TYPE”
‘Afghanistan’, ‘Country’
‘Andean Country Cluster’, ‘Cluster’
‘Asia-Pacific Region’, ‘Region’
‘Nepal Country Office’, ‘Country_Office’
‘Red Cross EU Office’, ‘Representative_Office’

With this approach, we would have five types: countries, clusters, regions, country offices and representative offices. The country offices and representative offices are important mainly for the surge deployments as IFRC personnel from these offices are sometimes deployed to other operations.

Let me know if I should clarify anything

Best regards,

Elsa

batpad:
Thank you @ElsaRaunioIFRC - that is perfect. I have made the required changes to the code. We shall let you know when this is available to see and edit in the Admin section.

batpad:
@szabozoltan69 I added the ability to filter districts by country into this PR, ref #458

PR #489, opened 2019-07-22
add is_enclave and record_type to district and country serializers

batpad:
@szabozoltan69 this adds the `is_enclave` and `record_type` data to the serializers so they can be consumed by the frontend.

Could you please look, test that it works okay and merge this? Thank you!


szabozoltan69:
Good idea, and I shall merge it in minutes...

PR #494, opened 2019-07-24
add new fields to field report model for new field report changes

batpad:
@szabozoltan69 I have added here the new fields that are required for the new Field Report changes, as per our discussion yesterday.

Please let me know this looks okay - happy to discuss it a bit more - am not sure if we need to add these fields in more places like the serializers, etc.

Thank you!


szabozoltan69:
This change is ok, Sanjay, just we should create the parallel change on frontend side also, because there are some conditions concerning the filled-up state of these fields. E.g num_missing:
```
views/emergency.js:155:    const numMissing = parseInt(get(report, 'num_missing')) || parseInt(get(report, 'gov_num_missing'));
views/field-report.js:168:                    <dd>{n(get(data, 'num_missing'))}</dd>
views/field-report.js:182:                    <dd>{n(get(data, 'gov_num_missing'))}</dd>
views/field-report-form/data-utils.js:180:    ['numMissing', 'num_missing'],
views/field-report-form/data-utils.js:432:    ['num_missing', 'numMissing'],
```

szabozoltan69:
Serializers also need some lift-up, e.g. MiniFieldReportSerializer...

PR #519, opened 2019-10-04
Changes for 3W

### LINKED ISSUE #514: 3w Projects: Allow user to select one Primary Sector of Activity + Many Secondary Sectors of Activity
Currently, there is a single field for the `sector` of a Project: https://github.com/IFRCGo/go-api/blob/master/deployments/models.py#L156

This needs to be split up into:

 - A `primary_sector` field which will refer to a single sector.
 - A `secondary_sectors` field which will refer to 0 or multiple sectors.

The idea being that a project will have a primary sector of involvement, but then also can have secondary sectors.

TODO:

 - [ ] Make changes to models
 - [ ] Make changes to API end-points (POST and GET)
 - [ ] Make changes to tests

cc @szabozoltan69 @frozenhelium @thenav56

### LINKED ISSUE #516: 3w Projects: Add field for Type of Operation
Add a field to the `Project` model under `deployments` for `operation_type` or so, based on these frontend changes: https://github.com/IFRCGo/go-frontend/issues/640#issuecomment-537422134 (need a new field to capture the new `Type of Operation` field on the backend)

cc @frozenhelium @thenav56

### LINKED ISSUE #532: 3w: Regional project
Add a separate field to capture multiple regions in the same project.

Suggested approach:
* Create a separate model which contains multiple regions
* Add a field to project form to capture the region group

thenav56:
Addresses **3W Project**

## Changes

- FIX #516
    - Add `operation_type` field to `Project`
- FIX #514
    - Change sector field to primary_sector
    - Add secondary_sectors field
    - Update Tests
    - Update Project Endpoints
- FIX #532
    - Add support for `regional project`.

* Add flake8 config
* Add PYTHONUNBUFFERED=1 to force server stdin, stdout and stderr to be totally unbuffered.

## Checklist 

- [x] #516
- [x] #514
- [x] #532
- [x] Additional Changes required by https://github.com/IFRCGo/go-frontend/pull/827


batpad:
This is all looking really good to me @thenav56 .

thenav56:
@batpad @frozenhelium I have pushed these changes https://github.com/IFRCGo/go-api/pull/519/commits/10e904cadfd9047b3e5d1e0ed11682b927a997c8 for regional project support. Can you guys look into it and let me know if it's what we want.

batpad:
> @batpad @frozenhelium I have pushed these changes 10e904c for regional project support. Can you guys look into it and let me know if it's what we want.

@thenav56 ah - my understanding of this was that it actually needs to be much simpler. Let me try and outline the use-case a bit and then my idea for the implementation:

 - Some projects cut across countries, and while each project should be associated with a country with individual reporting parameters, there needs to be some way to indicate that a project is part of a larger "regional project".
 - A "regional project" is just an aggregation of projects with a Foreign Key to it.

My proposal would be:
 - Call this new model `RegionalProject` (or `RegionProject`)
 - The only field this model has is a `name` (can include auto-fields like `created`, `modified`)
 - `Project` has a `ForeignKey` relation to `RegionalProject` - we can confirm with @guipizzini but I don't think a `Project` can be part of multiple `RegionalProject`s.

@thenav56 sorry for the misunderstanding here and the extra work - let me know if the above makes sense, happy to get on a quick chat to discuss if needed.

thenav56:
@batpad no worries, let's wait for confirmation from @guipizzini and I'll continue accordingly.

.dockerignore
```diff
@@ -0,0 +1 @@
+.tmp/
```
szabozoltan69: Great idea! I always had to move .tmp away and back before/after build manually. Thanks!

api/admin.py
```diff
@@ -5,6 +5,7 @@
 from django.utils.translation import gettext_lazy as _
 from django.utils.html import format_html_join
 from django.utils.safestring import mark_safe
+from django.db import models as dj_models
```
szabozoltan69: Where dj_models is used?

batpad:
This looks really good to me @thenav56 .

If @szabozoltan69 doesn't have any concerns, I'm :+1: to merge this and deploy to staging.

Thank you!

szabozoltan69:
Ok @batpad, feel free to merge + deploy. Do you remember the final "tag increase" rule of gitlab-CI to emerge the new docker image to docker hub? (And there is also a 2-steps authentication in the recent time for the azure login.) If there is any problem please write me and I resolve the deployment.

.dockerignore
thenav56: No problem, also we should probably look for other files which shouldn't be added to docker image.

api/admin.py
thenav56: Good find. I will remove this on the next PR if it's okay.

batpad:
@szabozoltan69 I will definitely need your help to deploy :-( - I'm still not setup on the deploy process. Happy to jump on another call next week if you think it's useful for me to get onboarded. For now, would really appreciate if you are able to deploy these changes to staging. Then, I can go ahead and merge the frontend code on Monday once you have deployed this to staging (no terrible rush if you are unable to get to it immediately).

szabozoltan69:
Hi @thenav56 @batpad – the branch is merged and deployed to staging environment.
Now it is up and running.
If you think I should put it to production also, give me a green light. (No problem if the frontend will arrive only later to production.)

PR #534, opened 2019-10-17
Implemented PER Draft deletion

GregoryHorvath:
Needed for the frontend change to be able to delete PER drafts. 

## Changes

* Implemented the API endpoint for the deletion which requires the `id` of the record.
* Added `id` to the serializer to send back for the frontend to use.

### DIFF
--- a/per/views.py
+++ b/per/views.py
@@ -429,6 +429,34 @@ class DelOverview(PublicJsonPostView):
 
         return JsonResponse({'status': 'ok'})
 
+def delete_draft(raw):
+    Draft.objects.filter(id=raw['id']).delete()
+    return
+
+class DelDraft(PublicJsonPostView):
+    def handle_post(self, request, *args, **kwargs):
+        u = None
+        richtokenstring = request.META.get('HTTP_AUTHORIZATION')
+        if richtokenstring:
+            try:
+                receivedtoken = Token.objects.get(key=richtokenstring[6:])
+            except:
+                return bad_request('User token is not correct.')
+            u = User.objects.filter(is_active=True, pk=receivedtoken.user_id)
+        else:
+            return bad_request('User token is not given.')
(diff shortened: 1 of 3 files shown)


per/views.py
```diff
@@ -429,6 +429,34 @@ def handle_post(self, request, *args, **kwargs):
 
         return JsonResponse({'status': 'ok'})
 
+def delete_draft(raw):
+    Draft.objects.filter(id=raw['id']).delete()
+    return
+
+class DelDraft(PublicJsonPostView):
+    def handle_post(self, request, *args, **kwargs):
+        u = None
```
batpad: Hmm is it not possible to get a reference to the user directly with `request.user` ? Having to get the token from the headers and manually look this up seems like it should not be required - i.e. the framework should handle this. Let me experiment with this a bit and maybe suggest a better method to get the user details from the request.

per/views.py
```diff
@@ -429,6 +429,34 @@ def handle_post(self, request, *args, **kwargs):
 
         return JsonResponse({'status': 'ok'})
 
+def delete_draft(raw):
```
batpad: Can we get this function to accept the `id` directly?

GregoryHorvath: Would this suffice the change you want: https://github.com/IFRCGo/go-api/commit/d7be6afd3bdfdc012c65e126fc7e991c04bcbeaa ?

batpad:
@GregoryHorvath I see that you're just following the same pattern as the other functions, so this is fine, I have approved the PR.

This way of getting the user data does seem a bit over-involved though, and also is leading to a lot of code repetition. @thenav56 can I ask you to look at this, and generally at the code inside `per/views.py` and we can discuss / perhaps make some tickets to refactor all these functions down the line to handle the getting of the user cleaner?

PR #549, opened 2019-11-25
Model changes for Early Warning Field Report

batpad:
Adds additional model fields to FieldReport to support new fields for the Early Warning Field Report type.

This is as per changes outlined here: https://github.com/IFRCGo/go-frontend/issues/833

@thenav56 can I ask for a review here?

cc @necoline -- this should complete all the backend work required to support the new fields and changes for the Early Warning Field Report.


batpad:
I added one additional field to the Field Report called `other_sources` for users to enter text about what the sources for their information is where they have marked `Other` as source.

PR #550, opened 2019-11-27
add a boolean for Event to be featured on the region page

batpad:
Towards https://github.com/IFRCGo/go-frontend/issues/846

This will allow admins to specify an Emergency to be featured on the Region page and Home-page separately, so an item can be featured on the region page, but not on the home-page.

It adds an additional filter on the Event end-point for `is_featured_region` to get events filtered by whether they are featured on the region page. It also adds this boolean to the serializer output for events.

cc @necoline

@thenav56 would appreciate a quick review, thanks!


szabozoltan69:
I deploy it to staging in 5 minutes...

PR #613, opened 2020-02-11
[Fix] Removed appealdocument__isnull from ingestion

GregoryHorvath:
Addresses https://github.com/IFRCGo/go-frontend/issues/968
@nanometrenat


batpad:
I'm not sure why we initially had `isnull=True` set - this looks good to me.

PR #621, opened 2020-02-20
Feature/audit trail

GregoryHorvath:
## Changes

* Added django-reversion and django-reversion-compare packages which handle creating snapshots of each record, being able to roll the changes back even if the record is deleted.
* Added logging for user login-logout attempts (with a failed status too).
* Added timestamps for Pending Registrations, to see when the admins approved (if they did) the registration, and to see how long the registration was pending.
* Added a "global" log page called 'Reversion difference logs' where all changes (and deletions also) can be seen.


batpad:
@GregoryHorvath I've started looking at this.

Overall, this looks good, but we do need to figure some things out before this can be deployed. Couple of issues I noticed:

 - The Recover Deleted Objects screen did not ever show me deleted objects, even after trying to delete some objects locally.
 - In the History view for an object and then clicking through a date, I did not actually see the previous revision. My guess is because we need to create initial reversions or so for all objects? We can discuss this.

The code overall looks good, I have some questions about some of the long `if else` logic statements, I'll leave some comments and we can discuss. This looks great and hopefully we can resolve these issues to be able to have it ready for testing soon.

api/receivers.py
```diff
@@ -0,0 +1,236 @@
...
+    'per.formdata': 'PER Form Data',
+    'per.nicedocument': 'PER Document',
+    'per.nsphase': 'PER NS Phase',
+    'per.overview': 'PER General Overview',
+    'per.workplan': 'PER Work Plan',
+    'registrations.pending': 'Pending registration',
+}
+
+
+def get_object_name(otype, obj):
```
batpad: Can you add some comments to explain what this function does and why, please? Then we can discuss if there's probably a way to do this more declaratively that avoids these long if statements with hard-coded values that can be hard to read / modify.

batpad:
In the Change History, the Changed From and Changed To values don't seem to be correct for me. Will do further testing of these things in a bit.

GregoryHorvath:
> * The Recover Deleted Objects screen did not ever show me deleted objects, even after trying to delete some objects locally.

It shows up for me, if you have the time let's have quick call to check it:
image

> * In the History view for an object and then clicking through a date, I did not actually see the previous revision. My guess is because we need to create initial reversions or so for all objects? We can discuss this.

Yes there is a command for it which creates an initial version of all objects, which I guess we would need to run manually after the deployment of this on prod: https://django-reversion.readthedocs.io/en/stable/commands.html#createinitialrevisions. I'm just not sure if we want to populate the logs with those init revisions.

api/receivers.py
GregoryHorvath: So this only gets the objects' `name` - which differs from model to model - for the global log page (which I called 'Reversion difference logs'). Since there is no switch-case statement in Python afaik, I had to go with the ugly and long if-elif-else solution.

Straight from reversion the model names come like `api.profile` but since reversion doesn't really cover the logging of deletions I had to create that manually and from that the model name comes like `profile` thus why I had to handle both. I agree that this is not the most maintainable version but I couldn't figure out how could I get a "globalized" model name and/or an object name for each different log in a prettier (standard) way.

I'd glad to fix this if you have any suggestions please don't hold it back 😄

GregoryHorvath:
@batpad Please run through 63891f5 if you have some free time, not so much changed so it should be quick.

Refactored `receivers.py`:
* Added `__str__()` to some Models missing it
* Removed the big `if-elif-else` block which was trying to figure out the name of the object/record for the log. Replaced it with `str(version)` which now it just gets them via `__str__()`
* Handled some key/null errors with the dictionaries
  * `<dict_name>.get(key, defaultValue)` is basically the usage of `get()` (basically it doesn't raise an error if it doesn't find the key)

If you find it good I propose to push this whole branch to staging for further testing, maybe together with PR: https://github.com/IFRCGo/go-api/pull/632 since it seems more and more urgent to have it working even on production (if we'll have the parameters for that), based on some e-mail chains.

PR #622, opened 2020-02-24
return fallback queryset, uncomment code

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/950#issuecomment-589101827

## Changes

 - Uncomments code to continue to return fallback queryset to not break preparedness tab on country pages

## Checklist 
 - Preparedness tab on country page frontend should not crash even when a country does not have PER / phase data.


GregoryHorvath:
@batpad Should I push this straight to staging? Or do we need this on production asap?

PR #628, opened 2020-03-10
Probable fix for updated records indexing

GregoryHorvath:
Addresses Elasticsearch indexing issue

## Issue description

It was trying to index new records as updated ones and it obviously failed if the records were just created and didn't have an index yet.

## Changes

Added a filter to the queries of `updated_*` records which checks if the `created_at` and `updated_at/modified_at/etc` datetimes were in the same minute, which querysets are used for the Elasticsearch indexing.

### DIFF
--- a/api/management/commands/index_and_notify.py
+++ b/api/management/commands/index_and_notify.py
@@ -694,6 +694,7 @@ class Command(BaseCommand):
             if mailcontents_of_fe:
                 send_followedevent_notifications(mailcontents_of_fe)
         else:
+            # ".annotate(diff...)" - To check if a record was newly created, we check if `created_at` and `updated_at` are within one minute of each other, since those values can be off by some miliseconds upon insertion.
             new_reports = FieldReport.objects.filter(cond1)
             updated_reports = FieldReport.objects.annotate(
                 diff = ExpressionWrapper(F('updated_at') - F('created_at'), output_field=DurationField())
--- a/api/management/commands/index_and_notify.py
+++ b/api/management/commands/index_and_notify.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timezone, timedelta
-from django.db.models import Q, Sum
+from django.db.models import Q, F, ExpressionWrapper, DurationField, Sum
 from django.core.management.base import BaseCommand
 from django.contrib.auth.models import User
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


GregoryHorvath:
https://github.com/IFRCGo/go-frontend/issues/989 is most possibly related.

batpad:
@GregoryHorvath can we talk through the logic of this a bit? Just seems to me like there should be a more straightforward way to do this.

batpad:
@GregoryHorvath this was a fantastic catch.

So, to just summarize the issue here:

 - For some reason (this possibly needs further investigation), when a new Emergency is created, the `created_at` and `updated_at` timestamps are not exactly the same - they can be off by a few miliseconds.
 - We were using a check to see if `created_at` was equal to `updated_at` and if it was, marking the record as a new record and indexing it accordingly, else marking it as `updated`. Of course, this was causing indexing to fail since `created_at` and `updated_at` would sometimes (?) not be the same, and the system would try and process a new record as an `updated` record and that would fail.

@GregoryHorvath - your fix looks really good to me, checking if `created_at` and `updated_at` are within a minute of each other. It would be really nice to dig a bit into if there's a way to ensure consistency of `created_at` and `updated_at` OR if there's other ideas / strategies on reliably checking if a record is newly created.

@thenav56 just checking in if this issue seems familiar and if you have any ideas for a simpler way of doing this.

@GregoryHorvath can we just add a comment in the code as to why these lines are required, something like:

```
To check if a record was newly created, we check if `created_at` and `updated_at` are within one second of each other, since those values can be off by some miliseconds upon insertion.
``` 

Really good work on setting up ElasticSearch locally and digging into this @GregoryHorvath - let's wait a bit to see if @thenav56 responds, else I'm :+1: to merging this.

thenav56:
@batpad Looks like field generates current timestamps separately so there will always be some **differences** when using auto_now/auto_now_add among **fields**.

One way to fix that will be to manually set `created_at` and `updated_at` using the same **datetime** by overriding the model `save()` method.

Something like this.
```python

class MyModel(models.Model):
    created     = models.DateTimeField(editable=False)
    modified    = models.DateTimeField()

        ''' On save, update timestamps '''
       now = timezone.now()
        if not self.id:
            self.created = now
        self.modified = now
        return super().save(*args, **kwargs)
```

But for now, @GregoryHorvath approach looks good as it works for all records.

PR #632, opened 2020-03-13
[FIX/CHANGE] E-mail notification sending via API

GregoryHorvath:
## Changes

* Reverted back the sending of Followed Events to its previous version, which means it uses the same method instead of a separate one.
* Rewritten `send_notification()` to use the API instead of using Python's smtplib.

## Checklist

Things that should succeed before merging:

- [ ] Need parameters for the API for the production environment from @szabolcsgajerifrc. (at the moment [2020-03-13] we have API parameters for the staging environment only) 
- [ ] Handle the response from the API.
- [ ] @batpad @thenav56 Review and/or comments on the code would be much appreciated. The interesting part is at `notifications/notification.py`.

The ENV variables come from the "env" file which is used for the deployment, where this is handled:

```
if [ $PRODUCTION == 1 ]; then
  ...
  export EMAIL_API_ENDPOINT="<API url for production>"
  export EMAIL_USER="<API username (email) for production>"
  ...
else
  ...
  export EMAIL_API_ENDPOINT="<API url for staging>"
  export EMAIL_USER="<API username (email) for staging>"
  ...
fi
```


GregoryHorvath:
Merging it for testing on the staging site, production parameters will come today most probably.

PR #636, opened 2020-03-17
vanity urls for events

geohacker:
Addresses https://github.com/IFRCGo/go-frontend/issues/1013

## Changes

* Add optional `slug` column to `Event` table
* Make DRF accept requests for both id and slug to get events

## Checklist 

- [x] Updated/ran unit tests
- [x] Updated CHANGELOG.md

## Release


api/models.py
```diff
@@ -243,6 +243,7 @@ class Event(models.Model):
     """ A disaster, which could cover multiple countries """
 
     name = models.CharField(max_length=100)
+    slug = models.CharField(max_length=50, default=None, unique=True, null=True)
```
batpad: this will need a `required=False` else the field is still required in the Admin interface.

GregoryHorvath: > this will need a `required=False` else the field is still required in the Admin interface.

Dunno about the `required` attribute, but `blank=True` would work for sure.

batpad:
@GregoryHorvath could you give this a test and review? And if all looks okay, we can deploy to staging.

What you should test:

 - Run migrations
 - In the admin, for an Emergency, try adding a `Slug`: eg. `test`
 - Then, fire up frontend talking to local db and navigate to http://localhost:3000/emergencies/test

You should see the Emergency with the slug `test`.

GregoryHorvath:
> @GregoryHorvath could you give this a test and review? And if all looks okay, we can deploy to staging.
> 
> What you should test:
> 
> * Run migrations
> * In the admin, for an Emergency, try adding a `Slug`: eg. `test`
> * Then, fire up frontend talking to local db and navigate to http://localhost:3000/emergencies/test
> 
> You should see the Emergency with the slug `test`.

Working awesome. Tested with trying to use the same `slug` for another Emergency and it gave me a warning that it's already in use. Perfect.

PR #642, opened 2020-03-23
Vanity url validation

geohacker:
Addresses https://github.com/IFRCGo/go-frontend/issues/1013#issuecomment-602530052

## Changes

* Validate the vanity url slug

## Checklist 

- [x] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -4,7 +4,7 @@ from django.utils import timezone
 from enumfields import IntEnum, EnumIntegerField, EnumField
 from .storage import AzureStorage
 from tinymce import HTMLField
-from django.core.validators import FileExtensionValidator
+from django.core.validators import FileExtensionValidator, validate_slug
 from django.contrib.postgres.fields import ArrayField
 from datetime import datetime, timedelta
 import pytz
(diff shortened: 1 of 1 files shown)


batpad:
@geohacker I think forcing the slug to be lower-case is probably the cleanest approach here. Should we add that to the `help_text` ? That the slug will always be made lower-case?

batpad:
This looks good to me. @GregoryHorvath if this looks fine, could you merge this to develop, and whenever possible, deploy `develop` to staging? Thank you!

GregoryHorvath:
@geohacker While this works fine, I feel like some users might still try to pass around frontend links like `.../emergencies/BigCatasthrope#details` and then wonder why they won't work. I tried to find a quick way to convert all incoming `slug` requests to lowercase but I haven't found one yet.

@batpad Should I merge it as it is right now or should we bother at all with mixed-case URLs?

teklal:
Hey folks, just took a look at this now on staging and having the text under the box is nice, thanks for that!

In my comment on https://github.com/IFRCGo/go-frontend/issues/1013 I mentioned also that the slug can't start with a number. I tested this now and got some weird behaviour:
image
image

A slug starting with a number does load an emergency page but not the correct one...?

Haven't investigated this now further but thought I'd flag.

GregoryHorvath:
> EDIT: Hmmm seeing that this PR is already merged I'll move this comment to the ticket
> 
> Hey folks, just took a look at this now on staging and having the text under the box is nice, thanks for that!
> 
> In my comment on [IFRCGo/go-frontend#1013](https://github.com/IFRCGo/go-frontend/issues/1013) I mentioned also that the slug can't start with a number. I tested this now and got some weird behaviour:
> image
> image
> 
> A slug starting with a number does load an emergency page but not the correct one...?
> 
> Haven't investigated this now further but thought I'd flag.

Awesome catch @teklal! I haven't noticed this but after checking it does the same for me (basically it gets the Emergency with the ID of the starting number).

geohacker:
Oh damn, that's right. I think it's trying match the ID. I'll add another validation to this same branch. Sorry about that! @teklal @GregoryHorvath

PR #652, opened 2020-04-03
Fix/create fr error handling

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1077


batpad:
@thenav56 @geohacker - can I ask one of you to take a look here? We have this issue that unfortunately we are only able to reproduce on production - https://github.com/IFRCGo/go-frontend/issues/1077 - adding some better error messaging seems to be the only way to debug this right now, since we seem to have failed at reproducing on staging or locally.

thenav56:
@batpad @GregoryHorvath Looks like you guys have already tried with the user data (from https://github.com/IFRCGo/go-frontend/issues/1077). I am also suspecting the error is related to hardcoded values.

Do you guys have asked for the API response from the users (from the chrome developer Network panel)? Since we are not logging serializer error in the server.

GregoryHorvath:
> @batpad @GregoryHorvath Looks like you guys have already tried with the user data (from [IFRCGo/go-frontend#1077](https://github.com/IFRCGo/go-frontend/issues/1077)). I am also suspecting the error is related to hardcoded values.
> 
> Do you guys have asked for the API response from the users (from the chrome developer Network panel)? Since we are not logging serializer error in the server.

@thenav56 The user checked the Chrome devtools afaik thus why they know the response was `400`, we didn't get any more details than that. Seemed to me if there was a `detail` property of errors on the frontend instead of the `error_messages` if `BadRequest` was raised.

PR #653, opened 2020-04-03
[WIP] fix incorrectly mapped actions in backend due to frontend bug

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/993

Work in Progress. Management command to create mappings of incorrectly saved actions to their correct counterparts

TBD: code to:
  - Fetch all field reports, by type, for specified date range
 - Find all actions associated with field report
 - If actions exist in the mapping for that Event type, re-map in the database

cc @GregoryHorvath @geohacker


batpad:
Closing as stale.

PR #656, opened 2020-04-06
make is_primary default false and not editable.

geohacker:
Addresses https://github.com/IFRCGo/go-frontend/issues/1008#issuecomment-608390658 

## Changes

* Migration to make is_primary on situation report types model to default
* Migration also has SQL to set selected report types is_primary to true

cc @batpad @GregoryHorvath


batpad:
:+1: looks good to me.

PR #661, opened 2020-04-08
Feature/merge emergencies

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1088, https://github.com/IFRCGo/go-frontend/issues/1010

## Changes

* Adds help text for `parent_event` field which tells you need to manually redirect connected Appeals (and such) if needed
* Fixes the frontend page not loading if an Emergency has both of a `parent_event` and a `slug`


GregoryHorvath:
@geohacker Please check if I used `**kwargs` the right way, I didn't have to use those parameters before.

geohacker:
@GregoryHorvath this looks really good to me! It has resolved the issue. One thing I noticed that's worth checking is that if you go to an emergency page that doesn't exist, it throws the same error. For example http://localhost:3001/emergencies/3570000#details. I think the EventDoesNotExist error needs to be caught and returned as a 404? cc @batpad

GregoryHorvath:
@geohacker Made some changes to handling errors there. Please recheck.

geohacker:
@GregoryHorvath Thank you! This looks good to me. I think the error I was seeing is coming from the front-end not handling it well. I made a PR to fix that https://github.com/IFRCGo/go-frontend/pull/1092

@batpad @necoline let me know if this makes sense.

GregoryHorvath:
> @GregoryHorvath Thank you! This looks good to me. I think the error I was seeing is coming from the front-end not handling it well. I made a PR to fix that [IFRCGo/go-frontend#1092](https://github.com/IFRCGo/go-frontend/pull/1092)

Oh I didn't check much on that part, nice catch! Merging and deploying this.

PR #678, opened 2020-04-24
Query objects instead of the database for actions

GregoryHorvath:
Could be helping with: https://github.com/IFRCGo/go-frontend/issues/1072#issuecomment-617427419


geohacker:
Nice catch @GregoryHorvath. I think this could definitely help!

PR #681, opened 2020-04-24
Fix/fr queries

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/1072#issuecomment-617427419

## Changes

 - Optimizes SQL queries on field report API end-point, reducing no of queries from 300+ to 9.

## Checklist 

 - [ ] Field Report API endpoint should work same as before, just much faster.

cc @GregoryHorvath


batpad:
@GregoryHorvath this should be good to merge to staging and test :+1:

PR #682, opened 2020-04-24
Django debug

batpad:
Adds the django debug toolbar while in development mode. This is especially useful to inspect SQL queries generated by an API end-point and optimize where necessary.

## Changes

 - Adds django-debug-toolbar to dependencies
 - Configure setup in settings and URLs

## Checklist 

- [ ] Verify debug toolbar works locally
- [ ] Verify it will not change anything in production

@thenav56 would be great if you could see and approve this one.

cc @GregoryHorvath

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -13,6 +13,8 @@ if BASE_URL == 'prddsgocdnapi.azureedge.net':
     BASE_URL = 'goadmin.ifrc.org'
 # The frontend_url nicing is in frontend.py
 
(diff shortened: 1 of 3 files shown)


thenav56:
@batpad Everything looks good.

Just out to point out, since we are using docker this might not work for local development.
```
INTERNAL_IPS = ['127.0.0.1']
```
You can get the host IP using `ip -4 addr show docker0 | grep -Po 'inet \K[\d.]+'`

PR #688, opened 2020-04-29
Feature/region project viz api

thenav56:
## Changes

- Add API For Regional 3W
- Add Secondary Tag COVID
- Update Project Filter
- Using Integer for Filter/Create/Update for enumfield for Project.

## Checklist 

- [x] https://github.com/IFRCGo/go-frontend/issues/1019

### DIFF
--- a/deployments/drf_views.py
+++ b/deployments/drf_views.py
@@ -18,15 +18,14 @@ from django.shortcuts import get_object_or_404
 from rest_framework import viewsets
 from reversion.views import RevisionMixin
 
+from .filters import ProjectFilter
 from .models import (
     ERUOwner,
     ERU,
     PersonnelDeployment,
     Personnel,
     PartnerSocietyDeployment,
-    ProgrammeTypes,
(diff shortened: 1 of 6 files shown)


batpad:
Also @thenav56 - not sure how much you've already done, but it would be great to ensure we've done some inspecting and tuning of the queries generated here. Again, if you think there is likely scope to improve query performance, add indexes, etc. please create a ticket -  I don't think it should hold us back, but it seems likely that spending some time with `EXPLAIN ANALYZE` and indexes may help provide some obvious speed-ups. Of course, if you feel that's already done and you have a good sense of query performance here, no need to ticket separately.

batpad:
@thenav56 - can you ping @GregoryHorvath if this should be merged to staging?

thenav56:
@batpad Sure will look into `EXPLAIN ANALYZE` to make sure the query is optimized.
And I will make sure to add tests addressing this on the next PR.

Issue #637, opened 2020-03-18
Allow users to choose if a change should trigger notifications


batpad:
**Is your feature request related to a problem? Please describe.**

This relates to https://github.com/IFRCGo/go-frontend/issues/1011 - breaking the part about allowing users to decide whether a change should trigger notifications into a separate ticket.

Often, users are spammed with notifications for every edit / change to something. We want users to be able to choose to not notify users on certain edits.

**Describe the solution you'd like**

When the user is editing an Emergency, they should have some option of ticking something like `Don't Notify` which, if they select, will not trigger sending of notifications.

@GregoryHorvath summed up the challenge here:

> The main issue here would be that we're not sending notifications on the go (haha pun), but checking every 5 minutes if there were any changes/new records and notifying the subscribers in a bunch. So even if one would uncheck the notify box if they're not waiting to the next 5th or 10th minute to check it back for the future or something, it wouldn't do much. This needs some brainstorming.

Let's use this issue to brainstorm and figure out how best to implement this feature where we can allow users to signify some edits as not triggering notifications.

cc @LukeCaley @GregoryHorvath

LukeCaley:
Feature to be prioritised due to volume of documentation related to COVID, i.e. technical guidance docs.  

Ideally there would be means from the back-end to select `notify subscribers **now**` and `notify subscribers in **daily update**` depending on urgency of the update.

There may be need for further design work on the notification email output here, cc @yellowumbrella-design

GregoryHorvath:
Had a call with @mmusori, and agreed that as a first approach we'd have a button at the top right of each Emergency/Event's page for superusers to send out a notification to the subscribers.

cc @LukeCaley

mmusori:
Pushing  Notifications.

Additional developments requested beyond the daily update for any changes to an emergency. 
There is a need to have a functionality to allow for pushing of a notification out immediately if there is important information that needs to be communicated. Potential solutions
1.	Have a button at top of a specific emergency page which is only visible to “super users/Regional IMs”. The button will read “Send Notification now”. So, if “important” changes have taken place a super-user will go in and “click/push” the button to immediately send a notification email.
 
2.		Have an additional (4th) save button at the bottom of a specific emergency. This button would be a  “Save and notify” button. This gives the authorisation to the person who modifies the emergency to either just save changes or to save changes and send a notification of the changes.


[...]

PR #689, opened 2020-04-29
Fix/notif times

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/637#issuecomment-608384908


api/management/commands/index_and_notify.py
```diff
@@ -467,7 +467,10 @@ def construct_template_record(self, rtype, record):
     def notify(self, records, rtype, stype, uid=None):
         record_count = 0
         if records:
-            record_count = records.count()
+            try:
+                record_count = records.count()
+            except:
+                record_count = len(records)
```
GregoryHorvath, line 473: @batpad @frozenhelium  @thenav56 There should be an easier way to distinct if the function gets a `queryset` or an `array`?

Since if it's an `array` then `.count()` requires an argument which determines what element to count **in** the `array`.
`if 'count' in dir(records): ...`  is not a feasible option since both have the `.count()` method, just used in other ways.

thenav56: @GregoryHorvath You can use python built-in function `isinstance`
eg:
```python
from django.db.models.query import QuerySet

# For Django Queryset
isinstance(your_qs, QuerySet)

# For Array
isinstance(your_qs, list)
```

PR #692, opened 2020-04-30
object_name CharField -> TextField

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1129

@thenav56 @batpad The weird thing is still that the `object_name` should not really get over just a couple 100 characters anyway since it's only using the `__str__()` method of the Models.
https://github.com/IFRCGo/go-api/blob/2f8c96884d4807b9031672c9ed6a9a79aeb6c177/api/receivers.py#L70 (and below)

But changed it into a `TextField` and tested the migration locally, all the previous data remained intact.


thenav56:
@GregoryHorvath I think the issue is because of **textfield** of FieldReport which is used for `__str__`.

https://github.com/IFRCGo/go-api/blob/2f8c96884d4807b9031672c9ed6a9a79aeb6c177/api/models.py#L616
https://github.com/IFRCGo/go-api/blob/2f8c96884d4807b9031672c9ed6a9a79aeb6c177/api/models.py#L802-L804

So your changes should be enough to handle this.

Issue #669, opened 2020-04-17
Create Django admin page for maintenance of whitelisted NS domains


teklal:
At the moment adding new email domains to the list of whitelisted goes usually as follows
- there is an email by a user to the IM email to tell about problems signing up
- someone from IM team contacts a developer to add the domain(s) to the lists
- there has to be a push to prod to make the emails work for registration

To make handling these domains easier in the future and avoid having to push things to production, it would be good to have a Django admin page for superusers to maintain the domains.

xref: https://github.com/IFRCGo/go-frontend/issues/556

teklal:
This is something that has been suggested by both Zoltan and @GregoryHorvath.

GregoryHorvath:
Deployed to staging.

@teklal

nanometrenat:
This was deployed to Prod just now.  Thanks!  Further tweak required please, to ensure it checks the entirety of the domain after the @ (i.e. not just a partial string compare).

e.g. asdfadsf@redcross.org.yyyyy

nanometrenat:
@teklal will leave it to you to close when happy with the interface - apparently the issue in the above comment was pre-existing ;)

GregoryHorvath:
> @teklal will leave it to you to close when happy with the interface - apparently the issue in the above comment was pre-existing ;)

@teklal @nanometrenat That issue should be solved with the next deployment (hopefully happening today).

GregoryHorvath:
This is on production already, closing.

teklal:
Was going to wait for the further tweak to be in prod before closing but all good - the major thing outlined in the ticket is already in prod and working. Thanks!

GregoryHorvath:
> Was going to wait for the further tweak to be in prod before closing but all good - the major thing outlined in the ticket is already in prod and working. Thanks!

Oh sorry for that I missed that's in this ticket still, it should be deployed by tomorrow morning though. Reopening until that.

teklal:
Closing as now everything mentioned in this ticket and comments is in prod. Thanks!

PR #709, opened 2020-05-12
Domain whitelist table instead of static list

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/669

Requires: https://github.com/IFRCGo/go-frontend/pull/1220


main/urls.py
```diff
@@ -125,6 +127,8 @@
     url(r'^api/v2/add_subscription/', AddSubscription.as_view()),
     url(r'^api/v2/del_subscription/', DelSubscription.as_view()),
     url(r'^api/v2/add_cronjob_log/', AddCronJobLog.as_view()),
+    url(r'^api/v2/event/(?P<pk>\d+)', api_views.EventViewset.as_view({'get': 'retrieve'})),
+    url(r'^api/v2/event/(?P<slug>[-\w]+)', api_views.EventViewset.as_view({'get': 'retrieve'}, lookup_field='slug')),
```
thenav56, line 131: @GregoryHorvath This will remove this action route https://github.com/IFRCGo/go-api/blob/254ae46fc63de967255023a290be6d592cf296d2/api/drf_views.py#L279-L281

Could you move this part below this?
https://github.com/IFRCGo/go-api/blob/254ae46fc63de967255023a290be6d592cf296d2/main/urls.py#L142

registrations/views.py
```diff
@@ -15,195 +16,31 @@
...
-        'ttrcs.org',
-        'vanuaturedcross.org',
-        'voroskereszt.hu' # no comma
+    allowed_list = [dom for dom in 
+        (
+            DomainWhitelist.objects
+                .filter(is_active=True)
+                .values_list('domain_name', flat=True)
+                .annotate(domain_name_lower=Lower('domain_name'))
+        )
```
thenav56, line 32: @GregoryHorvath We can also using filtering directly here instead of fetching all the data. Something like.
```suggestion
    is_allowed = (
        DomainWhitelist.objects
            .filter(is_active=True)
            .filter(domain_name__iendswith=domain)
            .exists()
    )
```
https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iendswith

registrations/views.py
```diff
@@ -15,195 +16,31 @@
...
     ]
+    
+    # Adding ifrc.org as a trusted domain always
+    if not allowed_list or 'ifrc.org' not in allowed_list:
+        allowed_list.append('ifrc.org')
+    # These 2 are needed for test_views.py
+    if 'arcs.org.af' not in allowed_list:
+        allowed_list.append('arcs.org.af')
+    if 'voroskereszt.hu' not in allowed_list:
+        allowed_list.append('voroskereszt.hu')
```
thenav56, line 42: @GregoryHorvath It will be a good idea to add fixtures to load previous static data and some for testing (remove this in production manually).

GregoryHorvath: @thenav56 I see what you mean, though those are part of the list anyway. It's just they were needed for the build to be even successful by CircleCI.

registrations/views.py
GregoryHorvath: Okies, should be simpler now: https://github.com/IFRCGo/go-api/pull/709/commits/f5126fe7b297fd01b360c376fcfeb6498b33ab88

main/urls.py
GregoryHorvath: Oh okay sorry this was a merge conflict and I thought it was just an accidental move or something. Fixed in a sec.

registrations/views.py
thenav56: Sounds good. 
PS: You can use directly `iexact` instead of annotating.

https://docs.djangoproject.com/en/3.0/ref/models/querysets/#std:fieldlookup-iexact

registrations/views.py
thenav56: Okay.
PS: You can use also `in` eg: ` if is_allowed or domain in ['ifrc.org', 'arcs.org.af', 'voroskereszt.hu']:`

GregoryHorvath: Wait sorry I missed to answer the real point of your first comment. So this is there just if anyone would happen to delete all the records in that table somehow it would still allow 'ifrc.org' for example so one could register that way. Same why I did a similar thing for the frontend validation.

PR #748, opened 2020-05-27
Field Report endpoint to return only public ones for not logged in users

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/pull/1286#issuecomment-633794448

@frozenhelium @thenav56 @batpad If you have any suggestions to improve the code don't hold back please. This was the only way I could see to make the logic work, but there might be better solution(s).

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -9,9 +9,11 @@ from django.http import Http404
 from django_filters import rest_framework as filters
 from django.contrib.auth.models import User
 from django.db import models
+from django.db.models import Prefetch, Q
 from django.utils import timezone
 from .event_sources import SOURCES
 from .exceptions import BadRequest
+from .utils import is_user_ifrc
 from .view_filters import ListFilter
 from .visibility_class import ReadOnlyVisibilityViewset
 from deployments.models import Personnel
@@ -246,10 +248,22 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
     )
     filter_class = EventFilter
 
+    def get_fieldreports_filter(self):
+        # Same logic at visibility_class.py
+        if self.request.user.is_authenticated:
+            if is_user_ifrc(self.request.user):
+                return Q()
+            else:
+                return Q(visibility=VisibilityChoices.MEMBERSHIP) | Q(visibility=VisibilityChoices.PUBLIC)
+        return Q(visibility=VisibilityChoices.PUBLIC)
+
     def get_queryset(self):
         if self.action == 'mini_events':
             return Event.objects.filter(parent_event__isnull=True).prefetch_related('dtype')
-        return Event.objects.filter(parent_event__isnull=True)
+
+        return Event.objects.filter(parent_event__isnull=True).prefetch_related(
+            Prefetch('field_reports', queryset=FieldReport.objects.filter(self.get_fieldreports_filter()))
+        )
 
     def get_serializer_class(self):
         if self.action == 'mini_events':
@@ -261,13 +275,18 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
 
     # Overwrite 'retrieve' because by default we filter to only non-merged Emergencies in 'get_queryset()'
     def retrieve(self, request, pk=None, *args, **kwargs):
+        filters
         if pk:
             try:
-                instance = Event.objects.get(pk=pk)
+                instance = Event.objects.prefetch_related(
+                    Prefetch('field_reports', queryset=FieldReport.objects.filter(self.get_fieldreports_filter()))
+                ).get(pk=pk)
             except Exception:
                 raise Http404
         elif kwargs['slug']:
-            instance = Event.objects.filter(slug=kwargs['slug']).first()
+            instance = Event.objects.filter(slug=kwargs['slug']).first().prefetch_related(
+                Prefetch('field_reports', queryset=FieldReport.objects.filter(self.get_fieldreports_filter()))
+            )
             if not instance:
                 raise Http404
         else:
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -248,22 +248,11 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
     )
     filter_class = EventFilter
 
-    def get_fieldreports_filter(self):
-        # Same logic at visibility_class.py
-        if self.request.user.is_authenticated:
-            if is_user_ifrc(self.request.user):
-                return Q()
-            else:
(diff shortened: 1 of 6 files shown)


batpad:
This looks great to me, @GregoryHorvath . No suggestions for improvement other than writing tests to test these permission cases. Can we ticket out writing tests to test all the permission logic around field reports, and then I'm good with merging this. :+1:

batpad:
@GregoryHorvath the way you've implemented this seems great. Just a couple little things we need to fix with the logic. We essentially need to make sure the same permissions logic is used as defined here: https://github.com/IFRCGo/go-api/blob/develop/api/visibility_class.py

I.e.

 - For non-logged in users, they only see PUBLIC field reports (visibility=3)
 - For users that are either `superuser` or have the `is_ifrc` permission: They should see ALL field reports.
 - For users that are logged in but are NOT `is_ifrc` or superusers: They should see all field reports EXCEPT those marked with visibility only to `VisibilityChoices.IFRC`

@GregoryHorvath let me know if this makes sense. I think some refactor could be possible, potentially breaking out the `is_ifrc` function into like a utils that takes a `User` object. But for now, just making sure this logic is applied and works to the field reports in the Event serializer is good.

thenav56:
@GregoryHorvath Also it would be a good idea to use prefetch here to avoid large database queries.
You can also add filters while prefetching.

This way you don't have to do anything in serializers.

eg: 
```diff
diff --git a/api/drf_views.py b/api/drf_views.py
index 9434bee..87201af 100644
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -1,3 +1,4 @@
+from django.db.models import Prefetch
 from rest_framework.status import HTTP_201_CREATED, HTTP_200_OK
 from rest_framework.generics import GenericAPIView, CreateAPIView, UpdateAPIView
 from rest_framework.response import Response
@@ -249,7 +250,10 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
         if self.action == 'mini_events':
             return Event.objects.filter(parent_event__isnull=True).prefetch_related('dtype')
-        return Event.objects.filter(parent_event__isnull=True)
+        field_report_filter = 'DEFINE YOUR FILTERS HERE'
+        return Event.objects.filter(parent_event__isnull=True).prefetch_related(
+            Prefetch('field_reports', queryset=FieldReport.objects.filter(field_report_filter))
+        )
 
     def get_serializer_class(self):
         if self.action == 'mini_events':
```
https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related

GregoryHorvath:
@thenav56 Changed the code to your suggested logic https://github.com/IFRCGo/go-api/pull/748/commits/77f4055e1722a87cedead48f49867b14d14c5cee and @batpad added a test in there https://github.com/IFRCGo/go-api/pull/748/commits/ed61d8e4d33587bb802df1f9d4193a890bb08549.

Please re-review.

batpad:
Thanks much @thenav56, did not realize you could apply the filters in the `prefetch` - that is indeed a whole lot nicer, and ideally this has the side-effect of speeding up queries a bit.

@GregoryHorvath the test seems straightforward and good to me. I'm :+1: on merging for QA unless @thenav56 has any further suggestions.

thenav56:
@batpad Yeah filters in prefetch is really good feature.

@GregoryHorvath  Everything looks good. I have one more suggestion here. I saw repeated queryset, so I think it would be nice if we define a function within the Event/FieldReport Model that accept's user and provides a filtered queryset.

Something like this.
```diff
diff --git a/api/drf_views.py b/api/drf_views.py
index 9434bee..569ad20 100644
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -249,7 +249,7 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
         if self.action == 'mini_events':
             return Event.objects.filter(parent_event__isnull=True).prefetch_related('dtype')
-        return Event.objects.filter(parent_event__isnull=True)
+        return Event.get_for(self.request.user).filter(parent_event__isnull=True)
 
     def get_serializer_class(self):
         if self.action == 'mini_events':
@@ -263,11 +263,11 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
     def retrieve(self, request, pk=None, *args, **kwargs):
         if pk:
             try:
-                instance = Event.objects.get(pk=pk)
+                instance = Event.get_for(request.user).get(pk=pk)
             except Exception:
                 raise Http404
         elif kwargs['slug']:
-            instance = Event.objects.filter(slug=kwargs['slug']).first()
+            instance = Event.get_for(request.user).filter(slug=kwargs['slug']).first()
             if not instance:
                 raise Http404
         else:
diff --git a/api/models.py b/api/models.py
index 2322cdd..ed89978 100644
--- a/api/models.py
+++ b/api/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.conf import settings
 from django.contrib.auth.signals import user_logged_in, user_logged_out, user_login_failed
+from django.db.models import Prefetch
 from django.dispatch import receiver
 from django.utils import timezone
 from enumfields import IntEnum, EnumIntegerField, EnumField
@@ -10,7 +11,7 @@ from django.core.validators import FileExtensionValidator, validate_slug
 from django.contrib.postgres.fields import ArrayField
 from datetime import datetime, timedelta
 import pytz
-from .utils import validate_slug_number
+from .utils import validate_slug_number, is_user_ifrc
 
 # Write model properties to dictionary
 def to_dict(instance):
@@ -313,6 +314,11 @@ class Event(models.Model):
         verbose_name = 'Emergency'
         verbose_name_plural = 'Emergencies'
 
+    @staticmethod
+    def get_for(user):
+        field_report_pretech = Prefetch('field_reports', queryset=FieldReport.get_for(user))
+        return Event.objects.prefetch_related(field_report_pretech)
+
     def start_date(self):
         """ Get start date of first appeal """
         start_dates = [getattr(a, 'start_date') for a in self.appeals.all()]
@@ -783,6 +789,15 @@ class FieldReport(models.Model):
     class Meta:
         ordering = ('-created_at', '-updated_at',)
 
+    @staticmethod
+    def get_for(user):
+        filters = models.Q(visibility=VisibilityChoices.PUBLIC)
+        if user.is_authenticated:
+            filters = models.Q(visibility__in=[VisibilityChoices.MEMBERSHIP, VisibilityChoices.PUBLIC])
+            if is_user_ifrc(user):
+                filters = models.Q()
+        return FieldReport.objects.filter(filters)
+
     def save(self, *args, **kwargs):
         # On save, is report_date or start_date is not set, set it to now.
         if not self.id and not self.report_date:
```

Issue #670, opened 2020-04-18
Provide a Rich Text Editor option for snippets


nanometrenat:
## Issue
Further to the below, and other conversations, we need to add a Rich Text Editor for Snippets 

- https://github.com/IFRCGo/go-frontend/issues/1011 
- https://github.com/IFRCGo/go-frontend/issues/931
- https://github.com/IFRCGo/go-api/issues/667

Particular driver right now is on Emergency pages, but Country and Regional pages also have snippets and it could be useful there too.

The Emergency Page snippets have

- a text box (not Rich Text) where users can paste in EITHER an iframe/URL link to embed e.g. a dashboard (this is the main use case and is critical to keep being supported) OR users can paste in html (this is the use case we're trying to help here - e.g. the Technical Guidance section of COVID-19 is html from another source copied and pasted in)
- an option to select an image for display (also need to keep this option)
- other options for visibility/position/tab/delete (these options also need to stay

What we'd like to have is a way for users to write Rich Text content (including ability to hyperlink to other sites, or to link to documents on GO) within the snippet so they don't have to create html elsewhere to paste in.

Hope this helps explain things but shout if further info required.

GregoryHorvath:
Just copying @teklal 's comment here from a duplicate of this and closing that issue: https://github.com/IFRCGo/go-api/issues/667#issuecomment-617654396

> - Surge alerts, the "Message" field https://prddsgocdnapi.azureedge.net/notifications/surgealert/add/ here the messages that are shared already most of the time have bulleted lists but they don't show on the frontend in the desired way
> - in field reports people sometimes add links to the "source" field so it would be good to be able to support the sharing of links there and having them properly show as links on the frontend too
> - snippets #670
>
> I'll add more when I think of other places.

GregoryHorvath:
@nanometrenat @teklal  Country, Region and Emergency Snippets are now RTE fields, deployed to staging.

nanometrenat:
Thanks @GregoryHorvath Unfortunately now I can't just copy and paste an iframe link into the snippet and have it display - if I just copy and paste directly then it shows as normal paragraph rather than as an iframe.  As per the description above iframes are our base use case for snippets.

In order to get an iframe to display I need to go into "Tools"/"Source code" and paste it there...
[...]

PR #749, opened 2020-05-28
Changed snippets into RTE fields

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/670

## Changes
- Country, Region, Emergency Snippets to Rich Text Editor (HTMLField - TinyMCE)

Will need extensive testing on staging but `<html>` snippets converted without a problem locally and were usable without any manual changes.

### Side notes
- Might need their width shrinked a bit for inlines (Snippets basically)

@thenav56 `makemigrations` made some migrations which affect `deployments` and `actionstaken`, please check on those:
- https://github.com/IFRCGo/go-api/pull/749/files#diff-287e79976830cab5ab15cc6d09369f09
- https://github.com/IFRCGo/go-api/pull/749/files#diff-70e34a7f98da04f5d79635390061c6ec


thenav56:
@GregoryHorvath Thanks for letting me know about those migrations.

This [migration](https://github.com/IFRCGo/go-api/pull/749/files#diff-70e34a7f98da04f5d79635390061c6ec) is due to the deletion of `null=True, blank=True` from project_districts field (my bad on forgetting to send those).

And this [migration](https://github.com/IFRCGo/go-api/pull/749/files#diff-287e79976830cab5ab15cc6d09369f09R14-R17) is due to the deletion of `blank=True` from your [commit](https://github.com/IFRCGo/go-api/commit/a573772a309ce1368bea60bec71dc39f88c9e34f).

Both of these migrations don't affect the database. So nothing to worry.

PR #753, opened 2020-06-02
Re-add FR visibility changes and tests

GergiH:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1118

We still need to change how we handle values showing up if they're not sent through the API by this endpoint.


szabozoltan69:
If I uncomment api/test_views.ps:test_field_reports_visibility, the test fails. Somehow the self.client.get('/api/v2/event/6') does not work (not even with .../1).
Maybe the solution (to create a getable endpoint) is in https://github.com/IFRCGo/go-api/blob/develop/TESTING.md - am I right?

Is there a possibility to keep the local test_test database? I tried pytest --reuse-db and pytest --create-db, but always drops test_test at the end.

batpad:
@szabozoltan69 I know you are out now, but will be great to remember to follow up on this one once you're back. Am going to close this as stale for now.

PR #754, opened 2020-06-02
Fix/email lowercase check

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1300

## Changes
- Password recovery now handles both lower- and uppercased letters
- Username recovery now handles both lower- and uppercased letters

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -12,7 +12,7 @@ from django.utils.decorators import method_decorator
 from django.contrib.auth import authenticate
 from django.contrib.auth.models import User
 from django.views import View
-from django.db.models.functions import Lower, TruncMonth, TruncYear
+from django.db.models.functions import TruncMonth, TruncYear
 from django.db.models import Count, Sum
 from django.utils import timezone
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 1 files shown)


api/views.py
```diff
@@ -383,7 +383,8 @@ def handle_post(self, request, *args, **kwargs):
             return bad_request('Must include an `email` property')
 
         try:
-            user = User.objects.get(email=body['email'])
+            req_email = body['email'].lower()
+            user = User.objects.annotate(email_lower=Lower('email')).get(email_lower=req_email)
```
thenav56, line 387: @GregoryHorvath You can use [`iexact`](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#std:fieldlookup-iexact) here directly.

```suggestion
            user = User.objects.get(email__iexact=body['email'])
```

api/views.py
```diff
@@ -410,7 +411,8 @@ def handle_post(self, request, *args, **kwargs):
...
```
thenav56, line 415: [Same](https://github.com/IFRCGo/go-api/pull/754/files#r433796168) here.

GregoryHorvath: Thanks again! Totally forgot about that you mentioned this in another place too. Done: bed7976

Issue #708, opened 2020-05-11
Simplify CSV Output for Event end-point


batpad:
Currently, the emergency API end-point, when returned as CSV, generates separate columns for every field report attached to the emergency. This is creating an inordinate and unmanageable amount of columns.

We should simplify the CSV output for the Emergency API so that it only returns summarized information for the field reports - perhaps just two fields:

 - Has Field Reports - simple `yes` or `no`
 - Field Report IDs - comma separated list of field report IDs

These changes should only happen on the CSV output, and we should leave the JSON output the same (we do this similarly to how we handle the Field Report serializer conditionally based on the output format).

@nanometrenat please let me know if this ticket is duplicated somewhere, could not find an existing ticket.

cc @GregoryHorvath is this something you might be able to take on?

nanometrenat:
Thanks @batpad that sounds great - I hadn't realised that it was simple to treat csv output differently to json - excellent solution.

Above 2 fields sound perfect.  The other thing that would be amazing (though would be an additional requirement, and may not be feasible, so not critical if too hard/risky), would it be possible to add an additional new field Has COVID Field Report ?  Would be really helpful for our dashboards but we can always do a join if not poss.  Many thanks

nanometrenat:
Interestingly, I did not realise that the "mini" event endpoint was even an option!  Interested to know how/where it is used:  https://goadmin.ifrc.org/docs/#event-mini_events

nanometrenat:
Following Monday's meeting, standing by for a proposed new schema for review - thanks!

kamicut:
So for the Events API, the included data that are larger objects are the following:
- Appeals
- Field Reports
- Countries

For countries, we can use the `database id` as well as an iso code or country name for legibility. I'm not too familiar with the structure of the Appeals and Field Reports but they each have a separate `aid` and `rid` that is different from the id.

Here's a sample output of the CSV with the current code:

[...]

PR #762, opened 2020-06-08
Minimize CSV fields in event export

kamicut:
Minimizes the number of fields in the CSV route for the event API 
addresses #708

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -70,6 +70,7 @@ from .serializers import (
     SnippetSerializer,
     ListMiniEventSerializer,
     ListEventSerializer,
+    ListEventCsvSerializer,
     ListEventDeploymentsSerializer,
     DetailEventSerializer,
     SituationReportSerializer,
@@ -255,7 +256,11 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
         if self.action == 'mini_events':
             return ListMiniEventSerializer
         elif self.action == 'list':
-            return ListEventSerializer
+            request_format_type = self.request.GET.get('format', 'json')
+            if request_format_type == 'csv':
+                return ListEventCsvSerializer
+            else:
+                return ListEventSerializer
         else:
             return DetailEventSerializer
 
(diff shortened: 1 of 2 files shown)


kamicut:
Not sure if this is the best way to go @batpad, because for countries the CSV will put the id of the countries instead of the name. Maybe we should revisit what fields should exist in the CSV output.

batpad:
> because for countries the CSV will put the id of the countries instead of the name.

@kamicut good catch. We may want to be able to define the field of the related model to use - eg. `id` in some cases, but `name`, etc. for others. For countries, my hunch is the best might be to use the ISO code or so.

@kamicut do you think if maybe possible to generate an example CSV and attach it to the ticket or so, and we can ask @nanometrenat to have a look and give feedback and what the ideal output for fields that were generating an arbitrary number of columns should be.

batpad:
@kamicut to write tests for this, you can look at something like this: https://github.com/IFRCGo/go-api/blob/develop/api/test_views.py#L56

In general, there's bits of the way we are setting up dummy objects in the db etc. for the tests that can be improved. But in this case, you can create a couple of events in your test with some dummy data, and then call with end-point with `format=csv` and check if the response is as expected.

You can run all the tests with `python manage.py test` (inside the docker container)

Let me know if it helps to look at the testing work-flow together.

api/drf_views.py
```diff
@@ -255,7 +256,11 @@ def get_serializer_class(self):
         if self.action == 'mini_events':
             return ListMiniEventSerializer
         elif self.action == 'list':
-            return ListEventSerializer
+            request_format_type = self.request.GET.get('format', 'json')
```
GregoryHorvath, line 259:
```suggestion
            request_format_type = self.request.GET.get('format', 'json')
            is_tableau = self.request.GET.get('tableau', 'false').lower()
            if request_format_type == 'csv' or is_tableau == 'true':
```

FYI - Couldn't figure it out yet how to work with this suggestion thingy. Consider :260 as changed to the above as well (last line). This is how I'll work with the Tableau WDC so I only need the CSV Serializer but with the JSON format. If you'll have the final version of how we're going to include the "name fields" into the generalized `CsvField()` please let me know.

batpad: @kamicut can you clarify on CsvField? I think for now we've just decided to go with more verbose Method fields and adding model methods? cc @GregoryHorvath

GregoryHorvath: @batpad @kamicut No worries I already merged and started doing it the new way, but any way it would be easy for me to change my current code if we'd change it again.

GregoryHorvath:
@kamicut Is this at a point that it can be merged or still WIP?

cc @batpad

PR #767, opened 2020-06-12
Fix/notifications visibility

GregoryHorvath:
## Changes
Separates the notification mail sending method into one which includes all Users who doesn't have superuser status or not in the 'IFRC Admins' group, and two those who do have either of those.

### Notes
There is a new `fix_types_for_subs` abstracted method, which feels pretty unnecessary if we only had the usable `RecordType`s and converted everyone's old subscriptions into those instead in the database. The whole `index_and_notify` file and related things could make use of a refactor/rewrite because this will only become a bigger mess later.

### DIFF
--- a/api/management/commands/index_and_notify.py
+++ b/api/management/commands/index_and_notify.py
@@ -83,16 +83,19 @@ class Command(BaseCommand):
         return countries, regions
 
 
-    def gather_subscribers(self, records, rtype, stype):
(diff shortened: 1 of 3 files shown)


GregoryHorvath:
This is deployed to staging and tested by myself and also partially with @teklal. Didn't find any issues in those tests.

PR #791, opened 2020-06-26
Minor/log location

GregoryHorvath:
Put `logger.log` into the folder alongside the other log files so it retains after a new deployment.

### DIFF
--- a/api/filehandler.py
+++ b/api/filehandler.py
@@ -1,0 +1,10 @@
+import logging
+import os
+
+
+class MakeFileHandler(logging.FileHandler):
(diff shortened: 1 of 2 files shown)


thenav56:
Hey @GregoryHorvath,  We would also need to update the deploy script https://github.com/IFRCGo/go-infrastructure/blob/develop/vms/init-api.sh#L65 to include volume for the new log directory.

GregoryHorvath:
> Hey @GregoryHorvath, We would also need to update the deploy script https://github.com/IFRCGo/go-infrastructure/blob/develop/vms/init-api.sh#L65 to include volume for the new log directory.

Oh yea sure, locally that's already done, I really need to start updating those files there.

PR #794, opened 2020-06-30
Comment out EnumField Labels

thenav56:
Comment out Labels from EnumFields.

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -54,12 +54,12 @@ class RegionName(IntEnum):
     EUROPE = 3
     MENA = 4
 
-    class Labels:
-        AFRICA = _('Africa')
-        AMERICAS = _('Americas')
-        ASIA_PACIFIC = _('Asia Pacific')
-        EUROPE = _('Europe')
-        MENA = _('Middle East North Africa')
+    # class Labels:
+    #     AFRICA = _('Africa')
+    #     AMERICAS = _('Americas')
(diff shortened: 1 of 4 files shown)


nanometrenat:
For xref purposes this was the symptom reported by users, that this PR fixed:
NB this is re csv exports (JSON unaffected)

> Some of the standard values in columns have changed, not the structure.
>  
> For instance:
> It used to be Dref, now it is DREF
> It used to be appeal, now it is Emergency Appeal
> It used to be Int, now it is International Appeal
> It used to be MENA, now it is Middle East North Africa

And this was the explanation given:
> As part of preparatory work toward making our APIs translatable, we accidentally made a change that caused ENUM values in the database to be represented by their verbose labels and not actual values in the CSV export. This was an accidental side-effect of marking strings for translations that caused us to change the structure of how we define ENUMs that caused this bug.
> 
> This did not get caught by either our manual QA process or our API test-suite, which currently only has tests for the JSON outputs, which were unaffected by this change. We are looking into improving our testing processes to ensure we catch these changes in the future before they affect downstream data consumers.
> 
> We are extremely sorry for the breakages this has caused on your end, but are very thankful for the catch. Rest assured, we will definitely be clear with communication when we do make purposeful changes to the API that may cause breakage for our users. In this case, it was an unintentional bug, and we are very grateful to your note that has helped us track it down and fix it.

PR #798, opened 2020-07-01
Removed RTE again

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/670, https://github.com/IFRCGo/go-frontend/pull/1356

@thenav56 https://github.com/IFRCGo/go-api/pull/798/files#diff-d238e1620f79a614231a52c09c6384e0


batpad:
@thenav56 would be great if you check that migration, it seems fine to me, likely the ENUM changes. But let's merge this to staging.

GregoryHorvath:
@batpad @thenav56 merging this then and we do another PR if necessary then, just to make progress with the pre-deployment process

thenav56:
> @thenav56 would be great if you check that migration, it seems fine to me, likely the ENUM changes. But let's merge this to staging.

Everything looks great. The migrations are because when commenting out the Enum fields I noticed the `Fact` wording was `FACT` previously so I have changed it back.

PR #803, opened 2020-07-01
Make Django Debug Toolbar work with docker-compose

GregoryHorvath:
- Added check for `DEBUG_INTERNAL_IP` env var
- Updated README with the additional steps


thenav56:
@GregoryHorvath Everything looks good.
But I would recommend to auto set the **DEBUG_INTERNAL_IP**. You could add a script **scripts/run_develop.sh** where **DEBUG_INTERNAL_IP** is calculated by the command you mentioned within the container and is passed to runserver.

GregoryHorvath:
> @GregoryHorvath Everything looks good.
> But I would recommend to auto set the **DEBUG_INTERNAL_IP**. You could add a script **scripts/run_develop.sh** where **DEBUG_INTERNAL_IP** is calculated by the command you mentioned within the container and is passed to runserver.

That's a nice idea, though I feared if it would affect production (or staging) in any way?

thenav56:
> > @GregoryHorvath Everything looks good.
> > But I would recommend to auto set the **DEBUG_INTERNAL_IP**. You could add a script **scripts/run_develop.sh** where **DEBUG_INTERNAL_IP** is calculated by the command you mentioned within the container and is passed to runserver.
> 
> That's a nice idea, though I feared if it would affect production (or staging) in any way?

Are we using django-debug-toolbar in production/staging? It's not recommended to use in production. It will impact the performance in production.

GregoryHorvath:
> Are we using django-debug-toolbar in production/staging? It's not recommended to use in production. It will impact the performance in production.

Nope, I meant just the INTERNAL_IPS array.

thenav56:
> Nope, I meant just the INTERNAL_IPS array.

Okay. Let's not include the `local_debug_ip` in `INTERNAL_IPS` when `DEBUG` is `false`.

GregoryHorvath:
@thenav56 Added the settings into `runserver.sh` (I was not sure how to do it properly with a new `.sh` file and this seemed it wouldn't be that big of a change, and adjusted the logic a bit how the `DEBUG` setting's value is calculated - staging also has `PRODUCTION_URL` set (but to dsgocdnapi...) thus why I needed another check there.

docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       AWS_TRANSLATE_ACCESS_KEY: $AWS_TRANSLATE_ACCESS_KEY
       AWS_TRANSLATE_SECRET_KEY: $AWS_TRANSLATE_SECRET_KEY
       AWS_TRANSLATE_REGION: $AWS_TRANSLATE_REGION
+      DEBUG_INTERNAL_IP: $DEBUG_INTERNAL_IP
```
thenav56, line 18: I will recommend we only use `DEBUG` mode in local development. cc: @batpad

main/runserver.sh
```diff
@@ -59,6 +59,7 @@ echo "export API_FQDN=\"$API_FQDN\"" >> $HOME/.env
 echo "export FRONTEND_URL=\"$FRONTEND_URL\"" >> $HOME/.env
 echo "export PRODUCTION=\"$PRODUCTION\"" >> $HOME/.env
 echo "export TEST_EMAILS=\"$TEST_EMAILS\"" >> $HOME/.env
+echo "export DEBUG_INTERNAL_IP=\"/sbin/ip route|awk '/default/ { print $3 }'\"" >> $HOME/.env
```
thenav56, line 62: My recommendation was to add a new script called `run_develop.sh` which be used to run a development version only. And add this logic there.
Something like this

**main/run_develop.sh** [add this to **docker-compose.yml** by replacing `python3 manage.py runserver`]
```sh
#!/bin/bash
export DEBUG_INTERNAL_IP=`/sbin/ip route| awk '/default/ { print $3 }'`
python3 manage.py runserver
```

main/settings.py
```diff
@@ -14,15 +14,23 @@
     BASE_URL = 'goadmin.ifrc.org'
 # The frontend_url nicing is in frontend.py
 
-INTERNAL_IPS = ['127.0.0.1']
+if PRODUCTION_URL is not None and 'prd' in PRODUCTION_URL:
+    DEBUG = False
+else:
+    DEBUG = True
```
thenav56, line 20: Are we setting `DEBUG=true` for all non-prod environments including staging? Enabling this might cause some security and performance issues.
cc: @batpad

thenav56:
@GregoryHorvath I have added some comments. Let me know if I am missing something.

main/settings.py
batpad: I completely agree. Saw that when going through things with @GregoryHorvath and it was unexpected - I would definitely approve of making this change to make sure DEBUG is only true on local environments.

GregoryHorvath: Right now `DEBUG=true` applies on staging as well because of the current logic, this is aimed to solve that.

docker-compose.yml
batpad: Agree. @GregoryHorvath do you think it's fine to make that change as part of this PR?

main/settings.py
batpad: @GregoryHorvath we should definitely have DEBUG be set to False in staging as well as production.

GregoryHorvath: @thenav56 Ah sorry, I now realized I simply had a brain-fart at the moment when testing this. The previous logic worked perfectly for staging as well. I'll roll that back. cc @batpad

GregoryHorvath:
Closing this because it got a bit out of hand and doesn't worth the effort at the moment, better start over with this later if needed.

PR #807, opened 2020-07-07
Update base docker image

geohacker:
This PR updates the base docker image from debian jessie to debian buster. It's essential to do this for fetching new packages for GDAL for using with geo django models.

## Changes

* Dockerfile uses a new image
* Updates some dependencies like Markupsafe and setuptools
* Remove pygraphviz as per conversation with @batpad 

## Checklist 

- [x] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,9 @@ EXPOSE 80
 EXPOSE 443
 
 RUN \
-	apt-get update; \
-	apt-get install -y nginx postgresql-client mdbtools vim tidy less gettext; \
-	apt-get install -y cron --no-install-recommends; \
+	apt-get update && \
+	apt-get install -y nginx postgresql-client mdbtools vim tidy less gettext && \
+	apt-get install -y cron --no-install-recommends && \
 	apt-get install -y binutils libproj-dev gdal-bin
 
 ENV HOME=/home/ifrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.3-jessie
+FROM python:3.6.11-buster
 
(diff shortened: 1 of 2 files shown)


batpad:
This looks fine to me @geohacker - `pygraphviz` was only used to be able to generate this graph of all the models as a management command. We can live without it, and I'm the only person who used that as far as I know. It will definitely be good to have some testing, but I think it's a good time to merge in changes like this, and I overall like the idea of updating our docker base image to a newer Debian.

@thenav56 @GregoryHorvath could you please take a look and approve if this seems all fine to you?

Dockerfile
```diff
@@ -1,12 +1,13 @@
...
 
 EXPOSE 80
 EXPOSE 443
 
 RUN \
 	apt-get update; \
 	apt-get install -y nginx postgresql-client mdbtools vim tidy less gettext; \
-	apt-get install -y cron --no-install-recommends
+	apt-get install -y cron --no-install-recommends; \
+	apt-get install -y binutils libproj-dev gdal-bin
```
thenav56, line 10: @geohacker Hey nothing major but could you also change `;` to `&&`
There might be cases where one of the line (7-9) throws error but last one doesn't.

GregoryHorvath:
@geohacker should I deploy this to staging then asap?

geohacker:
@GregoryHorvath I'll just fix the minor thing @thenav56 pointed out and then we can deploy to staging!

GregoryHorvath:
@geohacker deployed

nanometrenat:
Released this morning I am told

PR #808, opened 2020-07-07
Added last frontend login date

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/956

# Changes

- Added `last_frontend_login` to the 'User Profile' model and the CSV export
- Added the logic, (re)setting the date if a user logs in
- De-cluttered some of the logic
- Fixed most flake8 lint errors in the `view.py`


api/views.py
```diff
@@ -310,27 +317,30 @@ def post(self, request, *args, **kwargs):
...
             api_key, created = Token.objects.get_or_create(user=user)
 
-            # reset the key's created_at time each time we get new credentials
+            # Reset the key's created_at time each time we get new credentials
             if not created:
                 api_key.created = datetime.utcnow().replace(tzinfo=pytz.utc)
                 api_key.save()
 
+            # (Re)set the user's last frontend login datetime
+            user.profile.last_frontend_login = datetime.utcnow().replace(tzinfo=pytz.utc)
```
thenav56, line 341: FYI you can use Django timezone to get aware datetime directly.
```
now = timezone.now()
```
```suggestion
            user.profile.last_frontend_login = timezone.now()
```
https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/#naive-and-aware-datetime-objects

GregoryHorvath: Changed it on another place too and removed `pytz` since it's not needed after that. https://github.com/IFRCGo/go-api/pull/808/commits/3cc02486d12465d080533fbc09c7d712ac212933

PR #816, opened 2020-07-09
Added 'mini' serializer option to CountryViewset

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1233

Frontend PR: https://github.com/IFRCGo/go-frontend/pull/1377

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -145,7 +145,7 @@ class CountryViewset(viewsets.ReadOnlyModelViewSet):
             )
 
     def get_serializer_class(self):
-        if self.request.GET.get('mini', None) == 'true':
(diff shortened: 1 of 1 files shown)


api/drf_views.py
```diff
@@ -143,6 +145,8 @@ def get_object(self):
             )
 
     def get_serializer_class(self):
+        if self.request.GET.get('mini', None) == 'true':
```
thenav56, line 148: @GregoryHorvath You may want to use `lower` so that both `true`/`True` have the same meaning.
```suggestion
        if self.request.GET.get('mini', 'false').lower() == 'true':
```

PR #821, opened 2020-07-16
Added User Profile as inline to UserAdmin

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/956#issuecomment-659182089

## Changes
- Added 'User Profile' section (as inline, editable) to the 'User' - edit page

### DIFF
--- a/api/admin.py
+++ b/api/admin.py
@@ -24,7 +24,16 @@ from api.management.commands.index_and_notify import Command as Notify
 from notifications.models import RecordType, SubscriptionType
 
 
+class ProfileInline(admin.StackedInline):
+    model = models.Profile
+    can_delete = False
+    verbose_name_plural = 'User Profile'
+    fk_name = 'user'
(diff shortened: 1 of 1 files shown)


api/admin.py
```diff
@@ -24,7 +24,16 @@
 from notifications.models import RecordType, SubscriptionType
 
 
+class ProfileInline(admin.StackedInline):
+    model = models.Profile
+    can_delete = False
+    verbose_name_plural = 'User Profile'
```
thenav56, line 30: @GregoryHorvath Let's use lazy translation for verbose names.
```suggestion
    verbose_name_plural = _('user profile')
```

Note: https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#lazy-translation

> Just like you would do in verbose_name you should provide a lowercase verbose name text for the relation as Django will automatically titlecase it when required.

PR #822, opened 2020-07-17
prefetch_related for EventViewset query

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1412

Hopefully solves the CSV export slowness. Tested locally (left: without `prefetch` - right: with `prefetch`):
- 1m14s - 37s
- 56s - 24s
- 4m6s - 31s

As it seems it's pretty hectic but with `prefetch` it was at least twice as fast in every run.

cc @batpad @kamicut merging this to develop for staging testing

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -257,9 +257,15 @@ class EventViewset(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         if self.action == 'mini_events':
+            # FIXME: Why do we have an 'if' here when the (base) query is the same?
             return Event.objects.filter(parent_event__isnull=True).prefetch_related('dtype')
-        return Event.objects.filter(parent_event__isnull=True)
-        # return Event.get_for(self.request.user).filter(parent_event__isnull=True)
+        return (Event.objects.filter(parent_event__isnull=True)
+                             .select_related('dtype')
+                             .prefetch_related('appeals')
+                             .prefetch_related('countries')
+                             .prefetch_related('field_reports')
+                             .prefetch_related('regions')
+                             .prefetch_related('districts'))
 
     def get_serializer_class(self):
         if self.action == 'mini_events':


batpad:
:clap: @GregoryHorvath this *looks* good to me. Could you by any chance post a Query log before and after this change, and we can see if there maybe further query optimizations possible?

GregoryHorvath:
@batpad 
- before: ~272 queries (around 3-5 of it are something else)
- after: ~87 queries (same)

Measured through the Django Toolbar, sorry still no clue how I could've done it an other way :smile: (tried with `from django.db import connection` -> `connection['default'].queries` but it only gave me back the `SELECT` from the `api_event` table)

GregoryHorvath:
@thenav56 hey, I was trying some different approaches to this, suggested by multiple sources. The main thing was to manually construct/nest those prefetches, which in paper should've worked better than the current implementation, but it only resulted more queries and query-times for me.

> before: ~272 queries (around 3-5 of it are something else)
> after: ~87 queries (same)

**Do you have any ideas how to optimize the queries more?**

As of now we have:
https://github.com/IFRCGo/go-api/blob/74b5c88bc050119c30e2601239123080cfed8bb3/api/drf_views.py#L262-L268

and I tried something like this (leaving the last 2 as they were):
```python
        return (Event.objects.filter(parent_event__isnull=True)
                             .prefetch_related(Prefetch(
                                     'appeals',
                                     queryset=Appeal.objects.select_related('dtype')
                                                            .select_related('event')
                                                            .select_related('country')
                                                            .select_related('region')))
                             .prefetch_related(Prefetch(
                                 'countries',
                                 queryset=Country.objects.select_related('region')))
                             .prefetch_related(Prefetch(
                                 'field_reports',
                                 queryset=FieldReport.objects.select_related('user')
                                                             .select_related('dtype')
                                                             .select_related('event')
                                                             .prefetch_related('districts')
                                                             .prefetch_related('countries')
                                                             .prefetch_related('regions')))
                             .prefetch_related('regions')
                             .prefetch_related('districts'))
```

Actually if I only changed the `.prefetch_related('appeals')` to the second version it went from 87 queries up to 127 queries, but it went down to 92 with all the other changes. The more I feel to understand this the less it seems to be true.

thenav56:
@GregoryHorvath You don't have to use `.prefetch_related` for each prefetch. You should be able to do above with this
```python
        return (
            Event.objects.filter(parent_event__isnull=True)
            .select_related('dtype')
            .prefetch_related(
                'regions',
                Prefetch('appeals', queryset=Appeal.objects.select_related('dtype', 'event', 'country', 'region')),
                Prefetch('countries', queryset=Country.objects.select_related('region')),
                Prefetch('districts', queryset=District.objects.select_related('country')),
                Prefetch(
                    'field_reports',
                    queryset=FieldReport.objects.select_related('user', 'dtype', 'event').prefetch_related(
                        'districts', 'countries', 'regions', 'contacts',
                    )
                ),
            )
        )
```

Also after this, there was 11 queries only.
Also, note that prefetching can increase memory usages, so we should only include the attributes we are using (according to the serializer)

Also in the PR, I saw a `FIXME`, It's used to only prefetch `dtype` since other attributes are not used in `ListMiniEventSerializer`.

GregoryHorvath:
@thenav56 This is awesome! Thanks for clarifying for me how `prefetch` should work! Second PR will be up soon with this and updated the `mini` query as well to `select_related()`, although it only lowers the query count by 1.

Issue #812, opened 2020-07-09
Add testing coverage indicator


batpad:
Breaking out of https://github.com/IFRCGo/go-api/pull/806#issuecomment-655900648

Relates to https://github.com/IFRCGo/go-api/issues/797

As we work to beef up our automated test coverage, having a code coverage tool that calculates our test coverage on every commit will be useful. And it makes it much easier to enforce that new Pull Requests don't result in decreased test coverage.

I've used `Coveralls` in the past, would leave it here to @thenav56 to take a call on tooling.

It may not be realistic for us to aim for 100% test coverage as yet, but having an indicator of our test coverage and at the very least, ensuring it does not drop, will be an essential step toward getting there.

cc @thenav56 @gulfaraz 

(I know the `refactor` label is not quite accurate, but did not want to add a new label and I think most closely reflects the kind of issue this is, from a business perspective).

Issue #797, opened 2020-07-01
Add comprehensive test coverage for all API end-points


batpad:
**Is your feature request related to a problem? Please describe.**

As our API gets increased usage, it would be great to overhaul our approach to testing and work toward 100% test coverage of the API end-points. So far, we've got some tests to cover tricky bits of logic and aggregation, here being a nice example: https://github.com/IFRCGo/go-api/blob/develop/deployments/test_views.py

The patterns and organization for tests we use currently worked fine for the ad-hoc approach we've had so far for testing, allowing us to quickly write tests for tricky bits or for identified regressions. But it does not lend itself well to easily adding new tests for new features and end-points, or updating tests for changes, or coherently measuring test coverage. Changing the structure to allow for better code organization and developer friendliness, and having clear developer guidelines on adding and updating tests would go a long way toward this goal.

**Describe the solution you'd like**

I'd propose an overhaul of our API testing approach, and spending some time bringing consistency to how we write API tests. Below is a bit of a brain-dump of ideas, of roughly an approach I think makes sense, very happy to have more thoughts and suggestions here:

 - First, create a separate `tests` folder inside each app instead of the `test_views.py` single file. Within that, have a separate test file for each entity / API end-point being tested.

 - Create "factories" for each type of entity to easily populate the test database with fake data. This can be a bit of work to initially setup, but would then easily let us add many items to the test db and easily test all filters, etc. We would likely use something like `factory_boy` and `faker` or similar: https://medium.com/analytics-vidhya/factoryboy-usage-cd0398fd11d2

 - I'd love to use something like a "snapshot testing" approach, where our tests essentially:
   - Load test db with test data
   - Call each API end-point, with combinations of filters, limits, formats, etc.
   - Check each response against a "snapshot".
   - Have a good workflow to create initial snapshots, inspect when snapshots change, and update snapshots.

I have used Snapshot testing a fair bit in Javascript-land, but never with python / django. I found this package which seems to be the approach I'm advocating for: https://github.com/syrusakbary/snapshottest - I'd be surprised if there wasn't a recommended, consistent way to do this with Django + DRF, I'll do a bit more looking around.

@thenav56 would love to hear your thoughts and suggestions here.

@gulfaraz checking if this might be something you're able to take up - if you have the bandwidth and we can work together to setup the consistent structure and document test-writing, we can then split up work among all of us to cover all entities.

cc @GregoryHorvath @nanometrenat @LukeCaley

[...]

PR #829, opened 2020-07-25
feat: sample snapshot test and related doc update

gulfaraz:
Addresses #797 and #812  

## Changes

* Create sample snapshot-test for the Project model
* Write docs with instructions to add snapshot-tests

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md


batpad:
@gulfaraz this looks fantastic!

It would be really nice if we could extend the example test to test actual API calls.

Something like this, but using factories and a snapshot instead of testing individual properties: https://github.com/IFRCGo/go-api/blob/develop/api/test_views.py#L29

Ideally we'd be able to subclass our Test class both from the Snapshot testing TestCase as well as the APITestCase that's imported from `rest_framework` to add some helpers to test API endpoints (like self.client).

Ideally, snapshot testing would only be used for testing API endpoints, so it would be great if we can manage a simple example of that (can be extended later with testing with authentication, etc.). It's possible that it gets a bit weird trying to inherit from both APITestCase and the snapshot TestCase, happy to look at it further if things don't seem straightforward there.

Let me know if you're able to wrap this with some example of API testing - the structure of things and the factories, etc. look perfect, thank you so much for doing this.

batpad:
@gulfaraz this seems very close.

Ideally, we'd set things up so that we run the snapshot tests in addition to our current suite of tests. And will be great to figure out handling the client with auth, etc. a bit better. I'll need to spend a bit of time digging to really figure out the best approach here.

I'm okay merging this once we're sure it doesn't break running existing tests. However, it would be really nice to figure out cleanly mocking authorization, etc. @thenav56 would you have a bit of time to review and perhaps talk about this? This does seem almost there - we have factories to load mock data into the database (FANTASTIC), and we have an example call and snapshot that shows the idea behind making mock API calls and saving snapshots - it just seems like it could be wrapped a little neater into the django test runner ecosystem + integrated better with the Rest Framework Test classes - @thenav56 would be great to hear your thoughts on doing this better.

Thanks again for this @gulfaraz .

thenav56:
@batpad @gulfaraz. This looks good to me too.
And yeah I am also concern with the current TestCase and DRF/our helper functions.

I am also really new to snapshot testing, mostly done by providing data inside the test. But looking forward to learning more about this.

gulfaraz:
Thanks to @batpad's comment, I was able to create a simple `GET` request snapshot test on `/api/v2/project/` which returns an empty list.

When I attempted to create a project using `ProjectFactory.create()`, I was thrown this error `You need to provide an account name and either an account_key or sas_token when creating a storage service.`

From the docs, it's looked straightforward to perform [user auth in test](https://stackoverflow.com/questions/25534337/whats-the-proper-way-to-test-token-based-auth-using-apirequestfactory) and making [POST requests](https://stackoverflow.com/questions/25534337/whats-the-proper-way-to-test-token-based-auth-using-apirequestfactory). But the storage service error message blocked me from creating any reasonably useful tests.

I learned on a call with @batpad this morning that this `storage service` is a known bug outside the scope of snapshot tests or factory boy.

Snapshot tests are functional as the `GET project` test case seems to work as we expect.
I have not included an example of the factory usage but I've used nearly all their methods and the implementation works as expected. The only reference you would need to use factory boy is their [official docs](https://factoryboy.readthedocs.io/en/latest/#using-factories).

The test command first runs the default test runner and then also the snapshot tests. I did not have to do any additional config for this, django seems to pick up the snapshot tests just with [this setting](https://github.com/IFRCGo/go-api/blob/2856f9edff08f89f4c99fea12de8b91ab5854566/main/settings.py#L283) without ignoring the existing tests.

I've demonstrated the test coverage to @batpad which can be run using `docker-compose run --rm coverage_report` or `docker-compose run --rm coverage_html`.

These changes are documented in the [README](https://github.com/IFRCGo/go-api/blob/2856f9edff08f89f4c99fea12de8b91ab5854566/README.md) and [TESTING](https://github.com/IFRCGo/go-api/blob/2856f9edff08f89f4c99fea12de8b91ab5854566/TESTING.md) files.

It would be great if @batpad and @thenav56 would run the `docker-compose run --rm test` command on their local deployments to verify if the snapshot test works.

batpad:
@gulfaraz can confirm this works and the current tests run.

Am trying to figure out the issue with saving objects due to the error with the Storage Class, looking at https://stackoverflow.com/questions/22723122/mocking-a-custom-file-storage-backend

Do you have any code where you were trying to create objects using the factory classes that I can try to reproduce the error with the Azure Storage Class? I can then try some of these mocking / monkeypatching strategies to see if we can mock out the save method during tests.

gulfaraz:
> @gulfaraz can confirm this works and the current tests run.

Thank you for verifying.

> Do you have any code where you were trying to create objects using the factory classes that I can try to reproduce the error with the Azure Storage Class?

To see the error, add the below code to [deployments/tests.py](https://github.com/IFRCGo/go-api/blob/feat.snapshottest/deployments/tests.py) and run `docker-compose run --rm test`

    from deployments.factories import project

    def test_project_list_one(self):
        projectFactory = project.ProjectFactory.create()

        response = self.client.get("/api/v2/project/")
        self.assertEqual(response.status_code, 200)
        self.assertMatchSnapshot(response.content)

gulfaraz:
I've merged the fix for azure storage bug into this branch - #849

The error disappeared but the tests do not work as expected. They pass silently but not for the right reason.

In a second test case, I create a project using the ProjectFactory. Manually inspecting the creation of this project shows that a project has been created. But the project list API still returns an empty list.

    from deployments.models import Project

    print(Project.objects.all().count()) # gives 0
    ProjectFactory.create()
    print(Project.objects.all().count()) #gives 1

I need help figuring out why the list API does not show the new project.

batpad:
@gulfaraz this is looking great and seems to work for me!

While more examples are always welcome, I'd be comfortable merging this in along-with some documentation / guidance on writing snapshot tests. Am happy to help with that. Would be something along the lines of:

 - Use Factory / or write a separate Factory to create "fake" objects in the test database.
 - Test API calls with all the possible parameters, filters, etc. and create a snapshot of the response

A nice-to-have would be an example of easily creating multiple objects using the factories, to then run run list queries, filters, test pagination, etc. But if you need to jump onto other things, having figured out some of the nasty issues with the AzureStorage class and having this basic frame-work with factories working, am happy to create separate tickets and assign to have everyone pick up some API end-points and write tests based on the documentation and examples.

Nice work again, and very timely, it will be great to have this mechanism of maintaining snapshots as we're making changes to the API.

@thenav56 could you give this a look? It seems like it should be relatively straightforward to take these examples to generate more complex snapshot tests, loading more objects into the test db, passing different query parameters, etc.

batpad:
@gulfaraz lol, I just saw your comment above about things not working as intended after posting my comment. Let me take a closer look.

batpad:
>  But the project list API still returns an empty list.

@thenav56 could there be some other reason like a flag needed on the object or so before it appears in the List API for Projects?

gulfaraz:
@batpad Meanwhile, I'll work on improving the [testing doc](https://github.com/IFRCGo/go-api/blob/feat.snapshottest/TESTING.md) with the cases you mentioned.

The test cases with multiple objects are possible. I already know the syntax in my head but this creation issue is blocking me from exploring those scenarios. But we can definitely have examples for all sorts of filtering and ordering, it's very easy with factories.

If we're creating separate issues to write tests, I'd like to be assigned to the project APIs as those are the ones with which I'm most comfortable (yet). Maybe improving the doc can be a separate issue also with a lot more detail or as a part of this one - I'll let you decide.

thenav56:
> > But the project list API still returns an empty list.
> 
> @thenav56 could there be some other reason like a flag needed on the object or so before it appears in the List API for Projects?

@gulfaraz For project API, the queryset is filtered using **visibility**. If projects are created with **visibility** other then `public` then the user requesting should have permission according to the **visibility**.
The permission is related to logged in (member) or email ending with `@ifrc.org` (ifrc-user). For ifrc-user, the logic needs to be changed as mentioned here https://github.com/IFRCGo/go-frontend/issues/1312#issuecomment-672704352. For now, if you set the user's email ending with `@ifrc.org` then the project should return non-empty list.

gulfaraz:
Using the ProjectFactory, I created 50 projects,

    project.ProjectFactory.create_batch(size=50)

I manually inspected the **visibility** field to ensure that some projects were created with visibility as `public`.

I also created projects with users ending with `@ifrc.org`

    class UserFactory(factory.django.DjangoModelFactory):
        class Meta:
            model = get_user_model()
        username = factory.Sequence(lambda n: "user_%d" % n)
        email = factory.Sequence(lambda n: "user_%d@ifrc.org" % n)

I'm pretty sure @thenav56 meant that the user making the request to list all the projects needs to have an email ending with `ifrc.org`. I will look into how to set the test users email.

So far, the returned list is always empty. I can only confirm that the factory works as expected for one and many projects.

My next step is to understand how the list API is fetched and try to meet those constraints. I was able to create and read projects via the frontend so that's where I'm planning to start.

gulfaraz:
@thenav56 the empty list issue was fixed by explicitly setting the `visibility` value to `public` - Thanks for the pointer.

@batpad Now, the latest changes allows snapshot testing on the list projects API with zero and one projects.

Next, my plan is to write tests for the `/api/v2/project/` endpoints,
- [x] list/GET with 0 items
- [x] list/GET with 1 item
- [ ] list/GET with 2 items
- [ ] create/POST
- [ ] read/GET
- [ ] update/PUT
- [ ] delete/DELETE

After the above tests are written, I'll update the [testing documentation](https://github.com/IFRCGo/go-api/blob/8b0607f1810a080bf17421db57909b9fb8e7ce89/TESTING.md) to write additional tests.

batpad:
@gulfaraz this works great for me locally!

Will be great to add some simple documentation - I don't think it needs to be too detailed, the code is pretty self explanatory - your idea of running through an example with a simple model and then pointing to the existing tests sounds good to me.

@GregoryHorvath - could you go over this and let know if it makes sense and if you have any specific questions?

gulfaraz:
> @gulfaraz this works great for me locally!

Awesome. Thank you for verifying. 

> Will be great to add some simple documentation - I don't think it needs to be too detailed, the code is pretty self explanatory - your idea of running through an example with a simple model and then pointing to the existing tests sounds good to me.

I've created a [TESTING](./TESTING.md) file for documentation. It's already a part of this PR. Please review it's contents and confirm if the examples are sufficiently clear.

GregoryHorvath:
@batpad This looks awesome, thanks @gulfaraz !
I've ran through the code and the `.md` quickly, the only thing which is a little bit unclear to me still is what these "snapshots" really mean. But I'll probably get that when I'll see it in action.

batpad (approved):

batpad:
> I've created a TESTING file for documentation. It's already a part of this PR. Please review it's contents and confirm if the examples are sufficiently clear.

Sorry I missed that - that looks fantastic, thank you. I'm :+1: on merging this unless @GregoryHorvath has any concerns.

> the only thing which is a little bit unclear to me still is what these "snapshots" really mean. 

This is a very good question. Here's an article that explains the rationale behind this idea a bit, though the methodology used there is different: https://medium.com/better-practices/snapshot-testing-for-apis-using-postman-7f9f26295d6b

It's interesting that I think this style of Snapshot testing became popular with writing frontend UI tests, eg Jest: https://jestjs.io/docs/en/snapshot-testing

To try and explain a bit:

For most API tests you want to:

 - Populate the DB with some dummy data
 - Hit an API end-point querying the data, or POSTing new data, etc.
 - Check that the response returned from the API is as you would expect

You can see some of our earlier tests, we do something similar, but check only a subset of things about the response: https://github.com/IFRCGo/go-api/blob/develop/api/test_views.py#L48 - in that case, just that the API returns a 200, and that the count is correct. Ideally, we'd be checking in more detail that the objects and properties returned are exactly as we expect. But you can imagine, with the older style, checking through each property of the returned objects would be painful. Or if you maintained some JSON file that you compare it to manually, this would get painful to update by hand if the API changed slightly, etc.

This "Snapshot testing" mechanism simply gives us some easy tools to create and update those "snapshots" to compare responses to. So if you look at the `snap_tests.py` file, it essentially just stores the expected responses for your various tests, and gives you an easy mechanism to inspect what changed if something changed, and update the snapshots if the change looks as expected.

Hoping that gives a broad sense of it, of course, I think it will likely fully make sense once you start writing some tests for things :-) - happy to discuss this any time and figure out getting these API tests in over the next months.

batpad:
Going ahead and merging. This can just go with the next staging deploy @GregoryHorvath, since it only updates tests

Issue #743, opened 2020-05-25
Web Data Connector for GO - Tableau


guipizzini:
We need to build a Web Data Connector (WDC) to connect the GO API to Tableau. This should be a relatively straight forward task. In IFRC ESSN programme it took around 2 weeks to develop the WDC. Here a few links as a reference. No further specs are required on our end, we would like to have the whole API end points being pulled with the WDC.

https://github.com/tableau/webdataconnector

https://help.tableau.com/current/pro/desktop/en-us/examples_web_data_connector.htm

https://public.tableau.com/en-us/s/blog/2017/07/how-create-web-data-connector-beginners-approach

https://github.com/tableau/webdataconnector

batpad:
@guipizzini @frozenhelium - do we know whether this is something that @frozenhelium and team will be able to work on?

frozenhelium:
@guipizzini @batpad lets quickly discuss this sometime

guipizzini:
This has been paused upon internal agreements and discussions so that we can wait the improvements that are occurring on the API structure.

GregoryHorvath:
Had a discussion with @thenav56 @batpad that we could also do it totally separately as a Github site as it works for The Deep project, so it wouldn't affect the API development at all.

nanometrenat:
This link is to the KoBo WDC btw:  https://github.com/kobotoolbox/kpi/issues/2062

nanometrenat:
This link is to info about the KoBo WDC btw:  https://github.com/kobotoolbox/kpi/issues/2062

guipizzini:
We've tested the WDC with @GregoryHorvath and it is working fine and pending the deployment of other tickets. cc @batpad @LukeCaley

GregoryHorvath:
@guipizzini Deployed the GO API changes to staging. All https://ifrcgo.org/tableau-wdc/ endpoints should now be testable.

LukeCaley:
@GregoryHorvath @batpad could you update on the  deployment plan for this and whether there are any further developments / QA required? @guipizzini

batpad:
@LukeCaley this was deployed to staging, and was tested with the Tableau WDC "Simulator", but there seem to be some issues with actual Tableau. Hopefully this is quick to fix once @GregoryHorvath is back next week.

guipizzini:
[...]

PR #830, opened 2020-07-27
Serializers for Tableau WDC

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/743

WDC repo: https://github.com/IFRCGo/tableau-wdc

## TODO
- [ ] Confirm Field Reports' - Actions Taken stuff (fix in WDC repo too)


api/drf_views.py
```diff
@@ -278,7 +299,12 @@ def get_serializer_class(self):
         if self.action == 'mini_events':
             return ListMiniEventSerializer
         elif self.action == 'list':
-            return ListEventSerializer
+            request_format_type = self.request.GET.get('format', 'json')
+            is_tableau = request_format_type = self.request.GET.get('tableau', 'false').lower()
```
thenav56, line 303: You missed this.
```suggestion
            is_tableau = self.request.GET.get('tableau', 'false').lower()
```

api/serializers.py
```diff
@@ -39,43 +39,61 @@
...
         fields = ('name', 'summary', 'id',)
 
+
 class RegionSerializer(serializers.ModelSerializer):
     class Meta:
         model = Region
         fields = ('name', 'id', 'region_name')
 
 
+class CountryCsvSerializer(serializers.ModelSerializer):
```
thenav56, line 55: `CsvSerializer` sounds like it provides CSV. Maybe use `CountryFlatSerializer` or `CountryTableauSerializer`?
`CountryTableauSerializer` should be used, as this structure shouldn't be used/changed for go-frontend and provides scope into Tableau.
@batpad @GregoryHorvath What do you guys think?

api/serializers.py
```diff
@@ -198,10 +240,53 @@ class ListEventSerializer(serializers.ModelSerializer):
...
+    countries = serializers.SerializerMethodField()
+    dtype = DisasterTypeSerializer()
+
+    def get_countries(self, obj):
+        country_fields = {}
+        countries = obj.countries.all()
+        if len(countries) > 0:
+            country_fields['id'] = ', '.join([str(country.id) for country in countries])
+            country_fields['name'] = ', '.join([str(country.name) for country in countries])
+        else:
```
thenav56, line 262: We should avoid using `len(qs)`. Also instead of fetching all countries we should only fetch required field. Something like this.
```suggestion
        if countries.exists():
            country_fields['id'] = ', '.join([str(id) for id in countries.values_list('id', flat=True)])
            country_fields['name'] = ', '.join(countries.values_list('name', flat=True))
        else:
```

api/drf_views.py
```diff
@@ -202,18 +211,30 @@ class RegionSnippetViewset(ReadOnlyVisibilityViewset):
...
 
+
 class CountrySnippetViewset(ReadOnlyVisibilityViewset):
     authentication_classes = (TokenAuthentication,)
-    serializer_class = CountrySnippetSerializer
     filter_class = CountrySnippetFilter
     visibility_model_class = CountrySnippet
 
+    def get_serializer_class(self):
+        if self.request.GET.get('tableau', 'false').lower() == 'true':
```
thenav56, line 233: This is repetead many times. Maybe use a function `is_tableau`?
```python
def is_tableau(request):
    return request.GET.get('tableau', 'false').lower() == 'true'
```

api/serializers.py
GregoryHorvath: Thanks for the suggestion @thenav56 yea I was thinking about this too. It originated from the idea that I'd just use the `*CsvSerializer` for most endpoints and that would be good enough and it worked well for https://github.com/IFRCGo/go-api/pull/762. Only that most endpoints didn't have a different serializer for CSVs so where I needed it I just made one.

If you know a way to better extend out from the original serializers rather than copy-pasting them, changing 1 line and calling them `*CsvSerializer` please share :slightly_smiling_face: .
Something that comes into my mind is like (pseudo code):
```python
@overrides XYSerializer
class XYTableauSerializer():
    event = EventTableauSerializer()
    # and the rest is automatically there from the XYSerializer of course
```

cc @batpad

api/serializers.py
GregoryHorvath: Thanks! Commiting this for now, @kamicut tell me if we'd still want to keep your version so I can revert this back if needed.

api/drf_views.py
GregoryHorvath: https://github.com/IFRCGo/go-api/pull/830/commits/08e9d0fff77ace5fa3b8dec55192c668a795693f :+1:

api/serializers.py
thenav56: There might be issues if we override. For instance, if a non-flat field is added to the parent serializer then it will be also added to `*TableauSerializer`. Best will be the way you are doing right now. The reason I recommend using `*TableauSerializer` is that we will have the freedom to change and sync it with the web connector and it will not affect users who are using the CSV/Flat API directly.

api/drf_views.py
thenav56: Nice.
Quick Note: You can do `if is_tableau(self.request):` instead of `if is_tableau(self.request) is True:`.

api/serializers.py
GregoryHorvath: https://github.com/IFRCGo/go-api/pull/830/commits/74ea79738d0e3c6f306f1b4202915e4b3bb3c806

PR #833, opened 2020-07-28
Fix/remove old epi fields

GregoryHorvath:
## Changes
- Remove old EPI fields from `models`, `admin`, and added migration
- Remove related management commands which only needed to be ran for the data to be copied to the new fields


nanometrenat:
This PR doesn't have an issue reference but I believe it relates to https://github.com/IFRCGo/go-frontend/issues/1161#issuecomment-626391193

GregoryHorvath:
@nanometrenat Thanks, couldn't find the ticket. Reffing it.

Issue #737, opened 2020-05-20
[PROD] Emails not getting sent to email queue - registration - intermittent issue


nanometrenat:
## Issue
Spin off from MSM ticket SRV-21991 (more details re specific email addresses there)

Whilst on the whole the user registration process is working correctly (users get an email when they register, then they have to click on that email to verify their email address, then either they are activated - if on whitelist - or emails go to their verifiers).

However it was identified over the weekend that in at least one case an email was not sent to one of the verifiers (although an email WAS sent to the other verifier) so effectively that user was stuck in pending users until manual action is taken from an admin to activate them.

This ticket is to track investigations as to why the email was never sent to the internal email API that sends emails.

Whilst this particular missing email is the only one that investigations have already happened for (confirming it's not a user error), it is possible that there are wider intermittent email-sending issues elsewhere.  e.g. notifications. so once dev-investigations are done we will need to define what testing is required.

## Related feature
https://github.com/IFRCGo/go-frontend/issues/889

## Impact
Unknown - intermittent

## Severity
Complete blocker for those affected - they would need to contact IM for a manual update.

teklal:
Commenting here as also related to registration: there has been at least one occurrence (potentially more) of the _verification_ email not being sent to the user and the user hence not being able to verify and activate their account.

teklal:
Also tracking here as mentioned in the ticket that may be related to other emails as well:
I did not receive notification emails about surge alerts today although there have been 4 put on GO. 

So it is possible that this is a wider problem.

nanometrenat:
xref SRV-23022

GregoryHorvath:
Added `Notification GUID` "module" under the Notifications part on the Django Admin site. It saves all the emails' IDs which reach the sender API at least.
For most of the other cases more logging was added into the existing logs. If the recipients conversion fails a `Cronjob log record` warning will also be created with the address of one of the recipients.

All deployed to staging.

GregoryHorvath:
@karitotp @teklal 
## Notification GUID (Django Admin - under 'Notifications')
- `API GUID` - the ID coming back from the email sender API for the mail
- `Email type` - (has some additional info as well in most cases)
  - Field Report - new Field Report created
  - Emergency - new Emergency created
  - Validate account - new registration
  - Approve an account - to admins to approve a non-whitelisted account
  - Approved account successfully - self explanatory
  - Password recovery - forgotten password
  - Username recovery - forgotten usernam
[...]

PR #840, opened 2020-08-04
Re-send validation email, refactor to APIViews

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/737

## Changes
- Added `PublicJsonPostView` endpoint which requires the username and if its registration is pending it re-sends the validation email

Frontend PR: https://github.com/IFRCGo/go-frontend/pull/1490


api/views.py
```diff
@@ -435,6 +436,46 @@ def handle_post(self, request, *args, **kwargs):
         return JsonResponse({'status': 'ok'})
 
 
+class ResendValidation(PublicJsonPostView):
+    def handle_post(self, request, *args, **kwargs):
+        body = json.loads(request.body.decode('utf-8'))
```
thenav56, line 441: We should avoid using `PublicJsonPostView` and instead try to use DRF for any API views.

Something like this https://github.com/IFRCGo/go-api/blob/c1e11c4f382384ac2b0ba7048474b149f2eecd31/api/views.py#L51-L62

api/views.py
```diff
@@ -420,7 +421,7 @@ def handle_post(self, request, *args, **kwargs):
             return bad_request('Must include an `email` property')
 
         try:
-            user = User.objects.get(email__iexact=body['email'])
+            user = User.objects.filter(email__iexact=body['email']).first()
```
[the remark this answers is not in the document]
GregoryHorvath: Thanks, no clue how I missed that. Fix will be coming soon.

api/views.py
GregoryHorvath: All the other similar views use the same pattern, like `ShowUsername` or `RecoverPassword`. Should I also change those to use `APIView` too?

api/views.py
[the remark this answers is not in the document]
GregoryHorvath: dcaa397

api/views.py
GregoryHorvath: 4847a24

thenav56: We would need to refactor so that all API views are using DRF views instead of the custom `PublicJsonPostView`.
It will be cool to follow this for any new addition to the `go-api`.
We planned to do this right? @batpad  @GregoryHorvath

GregoryHorvath: @thenav56 Since these are public Views we don't need these for the `PublicJsonPostView`s, right?
https://github.com/IFRCGo/go-api/blob/c1e11c4f382384ac2b0ba7048474b149f2eecd31/api/views.py#L52-L53

The part here is what confusing me, though it doesn't seem that we use this anywhere atm:
https://github.com/IFRCGo/go-api/blob/c1e11c4f382384ac2b0ba7048474b149f2eecd31/api/views.py#L283-L286

thenav56: @GregoryHorvath Yeah we don't need them. The best way will be to just define empty **permissions_classses**
```python3
permissions_classes = []
```
Basically we should use `request.user` directly instead of `get_authenticated_user`, and for any custom logic, we should create a custom authentication class and add it to DRF **DEFAULT_AUTHENTICATION_CLASSES**. But yeah it isn't used anywhere so we don't need to worry about it.

per/views.py
```diff
@@ -1,70 +1,34 @@
...
+from django.http import JsonResponse
+from django.utils import timezone
+from rest_framework import authentication, permissions
+from rest_framework.views import APIView
 from .models import (
     Draft, Form, FormData, WorkPlan, Overview
 )
-from rest_framework.authtoken.models import Token
-
-def create_draft(raw):
```
GregoryHorvath, line 16: @batpad @thenav56  Removed most of these unnecessarily abstracted functions. Most of these can be found now inside the Views, ex. this one:
https://github.com/IFRCGo/go-api/pull/840/commits/b8b8e9ec3ed34b28d0a90879096284facb702d64#diff-8e4a20b18006b7a31a84250b0828251eR182

per/views.py
```diff
@@ -1,70 +1,34 @@
...
-
-        x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
-        if x_forwarded_for:
-            body['ip_address'] = x_forwarded_for.split(',')[-1].strip()
-        else:
-            body['ip_address'] = request.META.get('REMOTE_ADDR')
+from api.views import bad_request
+
+
+def get_client_ip(request):
```
GregoryHorvath, line 12: @batpad @thenav56 No clue if there's a nicer way to achieve this so I just made this a function from Zoltán's code. I don't even know why are we saving IPs here...

per/views.py
```diff
@@ -73,406 +37,300 @@ def handle_post(self, request, *args, **kwargs):
...
+            form.ended_at = ended_at
+            form.submitted_at = submitted_at
+            form.comment = comment
+            form.validated = validated
+            form.finalized = finalized
+            form.save()
+        except Exception:
             return bad_request('Could not change PER form record.')
 
-        if hasattr(form,'status_code') and form.status_code == 400:
```
GregoryHorvath, line 204: @batpad @thenav56 This felt redundant to me since we're already throwing a `bad_request` above if it fails. Is there any scenario where we might want to keep this in?

per/views.py
```diff
@@ -73,406 +37,300 @@ def handle_post(self, request, *args, **kwargs):
...
-            return bad_request('Could not change PER form record due to inner failure.')
+        # Update Form Data of the Form
+        if data:
+            try:
+                with transaction.atomic():  # all or nothing
```
GregoryHorvath, line 141: @batpad @thenav56 I hope I used `with transaction.atomic()` right. Also I guess we don't to save half-good Form Data so if one would fail, all should?

per/views.py
```diff
@@ -73,406 +37,300 @@ def handle_post(self, request, *args, **kwargs):
...
-        return JsonResponse({'status': 'ok'})
+                        form_data.save()
+            except Exception as err:
+                return bad_request('Could not change PER formdata record. {}'.format(err))
 
+        return JsonResponse({'status': 'ok'})
 
-class DraftSent(PublicJsonPostView):
-    def handle_post(self, request, *args, **kwargs):
-        u = None
```
GregoryHorvath, line 223: @batpad @thenav56 I guessed this part (repeating in most Views below) is just a custom way for doing:
```python
authentication_classes = (authentication.TokenAuthentication,)
permissions_classes = (permissions.IsAuthenticated,)
```
So I just switched them out.

per/views.py
```diff
@@ -1,70 +1,34 @@
...
-    form_data = FormData.objects.create(form= form,
-                            question_id     = raw['id'],
-                            selected_option = raw['op'],
-                            notes           = raw['nt'],
-                            )
-    form_data.save()
-    return form_data
-
-class FormSent(PublicJsonPostView):
-    def handle_post(self, request, *args, **kwargs):
```
GregoryHorvath, line 56: @batpad @thenav56 This didn't need authentication before but I added it now, so thus why the build failed: https://app.circleci.com/pipelines/github/IFRCGo/go-api/613/workflows/4c8458c8-22d6-491e-81e5-7f1e37800d85/jobs/2529

Should I remove the auth from this? It was weird to me why we allowed insertion without auth...

GregoryHorvath:
@batpad @thenav56 Also removed `PublicJsonRequestView` totally, tested the affected endpoints, all good on my side.

PR #844, opened 2020-08-06
use custom geo serializers for region, country and district

geohacker:
This PR introduces custom geo serializers for region, country and districts. It will ensure that other api end points that depend on the default serializers don't get burdened with the additional querying on bbox and centroid.

## Changes

* New geo serializers
* Region, Country and District list endpoints uses geo serializers
* Country endpoints now also return a `independent` field -- this will be used to determine whether the country needs to be available in the dropdowns etc.

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -57,11 +57,13 @@ from .serializers import (
     DisasterTypeSerializer,
 
     RegionSerializer,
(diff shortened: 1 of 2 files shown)


batpad:
This looks good to me and the geo-data no longer shows up for related objects.

Issue #825, opened 2020-07-22
[Local] Global File Storage configuration.


thenav56:
# Issue
Right now custom `AzureStorage` module is passed to each `FileField` for storing to azure storage. We need to implement global configuration so that `AzureStorage` is only used when credentials are passed to the application. If not, the files should be stored be in the local disk and also served from local disk.

Error when uploading in local.
![image](https://user-images.githubusercontent.com/7059255/88142359-18c27c80-cc15-11ea-9ec1-90c8c2e2c412.png)

# Solution
- Use [default file storage configuration](https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-DEFAULT_FILE_STORAGE) in Django. Only use `AzureStorage` if credentials are provided.
- Use sub directory to store media files locally `media/`
- Use Django storage plugin like [django-storages](https://django-storages.readthedocs.io/en/latest/backends/azure.html).

https://github.com/IFRCGo/go-api/blob/2a28e704c762e911d3e52fb9f5a149378696ced8/api/storage.py#L13-L217

# Checklist
- [ ] Don't define `storage=` argument in **FileField**.
- [ ] For **FileStorage**, use nested folder (using `MEDIA_ROOT`. https://docs.djangoproject.com/en/2.2/ref/settings/#media-root)
- [ ] Only use **AzureStorage** (django-storages) when credentials are provided. (Something like this https://github.com/the-deep/server/blob/develop/deep/settings.py#L274-L308.)

batpad:
@thenav56 this is turning out to be a blocker to the Snapshot testing PR: https://github.com/IFRCGo/go-api/pull/829 - the problem being that some of the models we are creating Factories for, have fields that have an AzureStorage class, and that's generating errors locally.

@thenav56 The solution you've outlined above seems to make sense - @gulfaraz do you think you maybe able to take a look ?  @thenav56 would you be able to provide some more guidance on "Use default file storage configuration in Django. Only use AzureStorage if credentials are provided."?

thenav56:
@batpad @gulfaraz 
This is the config used by the `AzureStorage` module in go-api.
https://github.com/IFRCGo/go-api/blob/edf4158d4386e7da87617a2d74fd8ed57cdebaa5/main/settings.py#L96-L101

Basically, if this config is not provided from environment variables then we use local storage. We can also add another env variable like `USE_AZURE_STORAGE` and use its value to switch between storage.
[...]

PR #849, opened 2020-08-20
Fix.azure storage

gulfaraz:
Addresses #825  

## Changes

* chooses file storage if azure storage details are not provided

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md


thenav56:
Hey @gulfaraz, Let's use [django-storages](https://django-storages.readthedocs.io/en/latest/backends/azure.html). Basically we would want to change the configuration on the global level (in **settings.py**) rather than in field level (**models.py**).

This should be the checklist
- [ ] Don't define `storage=` argument in FileField.
- [ ] For **FileStorage**, use nested folder (This seems to be done, but let's do this using `MEDIA_ROOT`. https://docs.djangoproject.com/en/2.2/ref/settings/#media-root)
- [ ] Only use **AzureStorage** (django-storages) when credentials are provided. (Something like in the example below.)

Eg: Configuration for using AWS S3 in Django https://github.com/the-deep/server/blob/develop/deep/settings.py#L274-L308

batpad:
@thenav56 do you think we could ticket your suggestions separately? I see what you mean where it would be better to use inbuilt Django Storages and follow a more native django pattern - it seems to me the current solution practically does a similar thing by setting the storage class based on the environment? This does seem to unblock both the snapshot testing and in general testing of file uploads locally, and I think @gulfaraz's time is likely better spent improving the snapshot tests right now than focusing on the Django Storages issue (which hopefully we can take up soon after this release) - @thenav56 does that sound okay, do you see any serious practical issues with the current approach?

@gulfaraz of course, if these changes seem straightforward, I'd be in favour of making them, but unless there's a strong objection from @thenav56 I feel comfortable moving ahead with this to unblock other things.

thenav56:
> @thenav56 do you think we could ticket your suggestions separately? I see what you mean where it would be better to use inbuilt Django Storages and follow a more native django pattern - it seems to me the current solution practically does a similar thing by setting the storage class based on the environment? This does seem to unblock both the snapshot testing and in general testing of file uploads locally, and I think @gulfaraz's time is likely better spent improving the snapshot tests right now than focusing on the Django Storages issue (which hopefully we can take up soon after this release) - @thenav56 does that sound okay, do you see any serious practical issues with the current approach?
> 
> @gulfaraz of course, if these changes seem straightforward, I'd be in favour of making them, but unless there's a strong objection from @thenav56 I feel comfortable moving ahead with this to unblock other things.

@batpad Sounds good. I have updated the issue https://github.com/IFRCGo/go-api/issues/825 with checklist.

thenav56 (approved):

PR #866, opened 2020-08-31
fix: modify migrations to accomodate azure storage

gulfaraz:
To avoid errors in the previously created migrations as an effect of using django storage #825 

## Changes

* Migration files with freed of their `api.storage` dependency


batpad:
@GregoryHorvath did you land up adding these migrations elsewhere? If not, this looks good to merge.

batpad:
@gulfaraz @GregoryHorvath do you know if we still need these?

gulfaraz:
@batpad this was to support the "django" way of implementing #825 

I see you've closed #825, in which case we don't need this PR anymore.

Issue #771, opened 2020-06-15
[PROD] https links for images uploaded as snippets so it doesn't give a browser warning


nanometrenat:
## Issue
Back end - On the edit Emergency page, snippets section, there's an option to upload an image so that the image can be displayed in the additional info (or other) tabs.

When I upload any image, the system gives it a URL - however the URL given by the system is an http URL rather than https - e.g. http://prddsgofilestorage.blob.core.windows.net/api/emergencies/3972/20200611_IFRC_COVID-19_Emergency_Allocations_as_at_8_June_with_annotation.png

![image](https://user-images.githubusercontent.com/33625192/84654866-31ea5600-af10-11ea-9a96-458987d285c1.png)

As noted by @batpad , this causes insecure content warnings when viewing the front end page (e.g. https://go.ifrc.org/emergencies/3972#actions)

There is no way for the user to edit the image URL to give it https - so need to either make it a https link, or omit the protocol  (e.g. //prddsgofilestorage.blob.core.windows.net/api/emergencies/3972/20200523_IFRC_COVID-19_Emergency_Allocations_as_at_15_May_manual_154.png ).  Omitting the protocol and it will use http:// or https:// based on the page it's being linked from.

You can tell if this issue is fixed because the browser will no longer show a little warning thing near the SSL lock in the browser bar.

Thanks!

GregoryHorvath:
The possible fix is deployed to staging.

PR #882, opened 2020-09-08
Changed blob urls to use https

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-api/issues/771

## Changes
Azure Storage (blob) URLs to use 'https' instead, this affects (API values too):
- Country - logo
- Country Snippet - image
- Region Snippet - image
- (Event/Emergency) Snippet - image
- Situation Report - document
- Appeal Document - document
- Databank > Key Document - file
- PER Document - document

### DIFF
--- a/api/storage.py
+++ b/api/storage.py
@@ -196,6 +196,7 @@ class AzureStorage(Storage):
         blob_url_args = {
             'container_name': self.container,
             'blob_name': name,
+            'protocol': 'https'
         }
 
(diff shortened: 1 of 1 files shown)


thenav56 (approved):

batpad:
This looks good to me @GregoryHorvath - I think this is an important enough fix to get in and test.

batpad (approved):

PR #905, opened 2020-09-24
update snapshots

batpad:
Fixes failing snapshot tests.

I was having a hard time inspecting the diff clearly to see what changed, but things look okay before and after.

@gulfaraz sorry I need this tests passing in a rush, so moving ahead with this merge, but let's look at this whenever you have some time to figure out why I needed to update the snapshots here.

cc @GregoryHorvath


batpad:
Oh wow, looked at the diff and these are all expected updates / changes to the API. This is fantastic and working as advertised!

PR #933, opened 2020-11-18
Geometries and Mapbox Tiling Service integration

geohacker:
This PR:
* [x] Move geom from country and district to new tables
* [x] Update the management commands to insert geometries to the new tables
* [x] Management command for preparing geojson to upload to mapbox
* [x] Upload tilesets to Mapbox


api/management/commands/prepare-admin-geojson.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
+import os
+from django.core.management.base import BaseCommand, CommandError
+
+class Command(BaseCommand):
+  help = "This command produces a countries.geojson and districts.geojson to be uploaded to Mapbox as a tileset. It is the source for all GO Maps."
+
+  def handle(self, *args, **options):
+    try:
+      DB_HOST = os.environ['DJANGO_DB_HOST']
```
batpad, line 10: you can probably get these a bit nicer by:

```
    db = settings.DATABASES[0]
    DB_USER = db.USER
    etc...
```

Probably a bit better to take these from settings than from the environ.

geohacker: Oh that's a great idea. Yes.

geohacker:
Note that to run the management command, a `MAPBOX_ACCESS_TOKEN` is required and this has to be added to the deployment by @GregoryHorvath.

GregoryHorvath (approved):

geohacker:
I think this is ready for merge @batpad.

batpad:
@geohacker this looks good to me.

There is some conflicts in the README.me - could you resolve and push, so we can merge?

I don't think we have plans to move this to a cronjob any time soon, especially because of the intermittent issues we've been seeing from the Mapbox API when uploading the tileset. Just to keep in mind that if we do move to a cronjob, it would be nice to follow the pattern in other cronjobs (see sync_molnix.py for example) to log outputs to a log file as well as publish the status of the cron to the admin. This looks really good to me! I'm :+1: to merge once the conflicts are resolved.

PR #936, opened 2020-11-24
Fix/region tabs

batpad:
Fixes some issues with Region Tabs implementation:

 - adds `additional_tab_name` to serializer for region and country
 - make `additional_tab_name` translatable
 - removes visibility permission levels for new snippet classes

cc @GregoryHorvath


batpad:
+ adds Country Cluster and National Society counts to the Region endpoint.

GregoryHorvath (approved):

PR #946, opened 2020-12-02
regenerate po files correctly from root of project

batpad:
Re-generates language .po files

cc @GregoryHorvath @thenav56 - it seemed to create new files in all folders, even where I don't think we changed strings, but I guess that's okay?

@thenav56 will be super great if you could eye-ball this and let know if it looks good. These were generated now by creating a `locale/` folder in the root and running it.


thenav56 (approved):

batpad:
@thenav56 I don't feel confident with what I'd done for the backend translations while you were away.

I'd feel better if we closed this one and you took a good look and made sure translation file generation all looked correct on the backend, and make a fresh PR? Am closing this - feel free to re-open and / or branch off from here - but it seems to me like regenerating all files from latest develop might be best?

Sorry about this.

PR #948, opened 2020-12-04
Fix national society count on region page in non-english languages

batpad:
Refs https://github.com/IFRCGo/go-frontend/issues/1621#issuecomment-738592798

cc @thenav56 - slightly interesting case, where I don't want the language localization to apply for a particular query - in general though, we should have a better way of telling whether a Country has a National Society than doing a check for empty string for the name, so this solution is likely okay for now.

cc @GregoryHorvath - please review / merge deploy when able, thanks!

### DIFF
--- a/api/models.py
+++ b/api/models.py
@@ -85,7 +85,7 @@ class Region(models.Model):
         return 'region-%s' % self.id
 
     def get_national_society_count(self):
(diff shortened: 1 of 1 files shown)


GregoryHorvath (approved):
@batpad Agreed that this seems to be a weird case. Would be nice if we could find a nicer solution later.

PR #956, opened 2020-12-08
ISO codes should be uppercase

geohacker:
ISO codes should be uppercase in the database. Fixes https://github.com/IFRCGo/go-frontend/issues/1681
This will require some frontend fixes.

@batpad @GregoryHorvath I've not been able to test this just yet as my db is locally broken because of out of order migrations.


GregoryHorvath (approved):

GregoryHorvath (changes requested):
Country doesn't seem to have an `iso2` but an `iso3` at least on my side. Other than this it looks good.

batpad:
@geohacker this looks good to me.

Could we add the `blank=True` to the `iso` and `iso3` fields as part of this PR as well, to make them non-required in the admin?

GregoryHorvath (approved):

Issue #820, opened 2020-07-15
[PROD] Change wording of tickbox for hiding field report figures on emergency page


teklal:
## Issue
On emergency pages with field reports linked to them, there is the possibility on the Django admin to hide the numeric details (called Emergency overview on the emergency page frontend). 

section in question on field report page
![image](https://user-images.githubusercontent.com/52411237/87546345-b3690b80-c6a9-11ea-9e0c-637a4746d7df.png)
on emergency page
![image](https://user-images.githubusercontent.com/52411237/87546382-c380eb00-c6a9-11ea-8334-8cea2fdc9c9f.png)

This tickbox currently says "Hide attached field reports?" which may be misleading as the only thing it actually hides is the aforementioned numeric details. The field report itself, situation overview, map and contacts that are pulled from the linekd field report will still show on the page even when this box is ticked.

![image](https://user-images.githubusercontent.com/52411237/87546230-8157a980-c6a9-11ea-9fbc-ad76682f2d0a.png)

## Expected behaviour
The tickbox in the backend to say "Hide field report numeric details" or something along those lines.

## Related feature
https://github.com/IFRCGo/go-frontend/issues/1397

nanometrenat:
@batpad  is this ready for QA?

batpad:
Sorry, this did not fully get through in 4.3. Moving to 4.4.

karitotp:
The tickbox was changed and now it is "Hide field report numeric details". Also, it is working as is expected. cc. @batpad

PR #962, opened 2020-12-11
Changes to Field Report Model

kamicut:
- Change wording of tickbox for hiding field report figures on emergency page #820 
- Add option of hiding map on emergency page #1397

This might need some translation fixes before merged @batpad, but wasn't sure about how to generate the translations or what the process is there.


batpad:
@kamicut for tests, you should be able to regenerate snapshots and check the diff looks expected, see https://github.com/IFRCGo/go-api/blob/develop/TESTING.md

For translations, that is a very good question. We need to get better with our process around translations. Right now, we have not been merging translation strings for each PR with the PR, as I imagine we should do in the future.

For now, let's merge this without translations and I can merge this into this PR alongwith other translation updates: https://github.com/IFRCGo/go-api/pull/946

batpad (approved):

PR #994, opened 2021-01-21
Allow login with email besides username, validations

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1663

## Changes

- Added new Authentication Backend to also check against `email` when authenticating. This is also a workaround to have a `CustomUser` model (`username` is a required field by default in Django's `User` model) which would mean we would need to migrate existing User data (and their relations) to this new model.
- Added validation for password reset/change, so now users should get an error message whenever they try to use an invalid password. Currently even invalid passwords are getting saved, yet the application won't let them login with that password.
- Removed `username` / added `email`, from/to required fields where they were needed.
- New registrations now automatically fill in the `username` with the `email`.
- Adjusted New Registration tests to follow this scheme.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -10,6 +10,7 @@ from django.views.decorators.csrf import csrf_exempt
 from django.utils.decorators import method_decorator
 from django.contrib.auth import authenticate
 from django.contrib.auth.models import User
+from django.contrib.auth.password_validation import validate_password
 from django.conf import settings
 from django.views import View
 from django.db.models.functions import TruncMonth, TruncYear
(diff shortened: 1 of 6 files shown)


api/views.py
```diff
@@ -272,15 +273,14 @@ def post(self, request):
...
 
         if username is None or password is None:
-            return bad_request('Body must contain `username` and `password`')
+            return bad_request('Body must contain `email/username` and `password`')
 
         # Get the case-correct username for authenticate()
-        casecorr_uname = User.objects.filter(username__iexact=username).values_list('username', flat=True).first()
-        if casecorr_uname is None:
-            return bad_request('Invalid username or password')  # definitely username issue
+        # casecorr_uname = User.objects.filter(username__iexact=username).values_list('username', flat=True).first()
```
batpad, line 279: can we drop the commented lines of code before merging, @GregoryHorvath ?

GregoryHorvath: Done

PR #1007, opened 2021-02-11
Add iso3 filter to aggregate endpoint

GregoryHorvath:
This is mainly for FDRS but we could also use it for the header key figures on the frontend.

@thenav56 Merging this for testing, but could you please check this part and tell me if there's any better way to handle such conditional filtering?
https://github.com/IFRCGo/go-api/pull/1007/files#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R151-R157

Since after the `annotate` we have aggregated values only so filtering after doesn't seem to be an option.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -6,8 +6,6 @@ from rest_framework import authentication, permissions
 
 from datetime import datetime, timedelta
 from django.http import JsonResponse, HttpResponse
-from django.views.decorators.csrf import csrf_exempt
-from django.utils.decorators import method_decorator
 from django.contrib.auth import authenticate
 from django.contrib.auth.models import User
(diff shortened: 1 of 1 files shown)


thenav56:
@GregoryHorvath Annotate logic looks fine to me. Not sure if there is a better way to do it. But
- We should use enum variables instead of the raw value. (eg: use AppealType.DREF instead of 0)
- There are `Count(`/`Sum(` in annotating which looks redundant to me.
- Remove duplicate annotate/aggregate statements by only changing the Appeal queryset for iso3. Basically, update/change the queryset before the annotate/aggregate statement.

PR #1032, opened 2021-03-11
filter national society counts for regions to only include independent=True

batpad:
Only include countries with independent=True in National Society counts for regions.

cc @jhenshall 

@GregoryHorvath will be great to get this on staging.


GregoryHorvath:
@batpad deployed.

PR #1045, opened 2021-04-08
only show country from of active deployments

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/1836 

## Changes

Fixes bug where we still returning Countries From for deployments that were no longer active. Now we should only return countries from for active deployments per Emergency.

@szabozoltan69 could you review and merge to develop and deploy whenever is possible? Thanks!


batpad:
@szabozoltan69 bumping this -- possible to merge and deploy to staging soon?

szabozoltan69:
Of course, it can be deployed soon.

Issue #1026, opened 2021-03-03
Track usage of the go-api


JonathanGarro:
# Track usage of the go-api
My assumption is that **currently we do not track any usage**. If we already do then parts of the proposed tasks can be considered to be in-progress. I've split the effort into the following packages such that they can be picked up one after the other. However for the sake of efficiency I recommend we decide beforehand our goal tier.

## The tiers of commitment are,

1. **Basic**

We use a 3rd party service such as Microsoft Application Insights or Google Analytics. This is sufficient to feel the pulse on usage of start up apps.

### Features
1. Count number of times a certain endpoint is called.
2. Parameters with which these endpoints are called.
3. Count the number of errors or failures.
4. These metrics will be available in the dashboard of the 3rd party service. We would have to design the charts we consider important.

![Screen Shot 2021-03-03 at 11 21 21 AM](https://user-images.githubusercontent.com/8890661/109838588-4d1f0200-7c14-11eb-87b6-59165338be62.png)

A example Microsoft Application Insights dashboard using metrics gathered from a front end app.

### 40 hours of effort if we already know which metrics are important.

2. **Light**

We build an in-house analytics service. This gives us full control of the data but *with great power comes great responsibility*. Responsibilities include building, running and maintaining a separate analytics server and database independent of Go.

### Features

1. Includes all the [**Basic** tier features]().
2. 3rd party services don't use all the collected data but an approximation to be able to scale. If we control the data, we can use actual numbers instead.
3. We can build a dashboard on the go-frontend itself instead of navigating to a 3rd party website. *There are [ways to embed 3rd party charts](https://developers.google.com/analytics/devguides/reporting/embed/v1/) but I don't recommend that approach.*
 
![Screen Shot 2021-03-03 at 11 22 30 AM](https://user-images.githubusercontent.com/8890661/109838643-57d99700-7c14-11eb-9d5a-493943021f9a.png)

An example of a custom dashboard. [Source](https://demo.baremetrics.com/)

### 240 hours of effort + [Basic tier effort]().

3. **Expert**

We build analytics as a feature to our end users. We provide users access to the API via OAuth tokens generated using the Go interface. Each user can generate multiple tokens which get tracked separately. Each token comes with a dashboard showing how the API is being used through that token.

### Features

1. Includes all the [**Basic** tier features]().
2. Includes the [Light tier features](). The global dashboard is optional but does not really save significant effort — ~20 hours.
[...]

PR #1057, opened 2021-04-22
feat: analytics logging for incoming requests

gulfaraz:
Addresses https://github.com/IFRCGo/go-api/issues/1026 

## Changes

* Log incoming API requests to Azure
* This PR adds a new environment variable
`APPLICATION_INSIGHTS_INSTRUMENTATION_KEY=fa128365-fb09-4568-b944-b05e38b6b94c`

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


batpad:
This looks really good to me, @gulfaraz !

@thenav56 do you mind giving this a review?

@szabozoltan69 once this is merged, we will need to add this key to the environment variables -- @gulfaraz should we setup different keys for staging and prod? or leave it undefined on staging?

Also, @gulfaraz can it be run locally without a key?

szabozoltan69:
So I added (in both staging and prod scripts) to init-api.sh:
echo "APPLICATION_INSIGHTS_INSTRUMENTATION_KEY" >> .tmp/env
and to .env:
export APPLICATION_INSIGHTS_INSTRUMENTATION_KEY='...'
Now comes the merge and deployment... (in 10 minutes it is on staging)

szabozoltan69:
Deployed to staging

gulfaraz:
> can it be run locally without a key? or leave it undefined on staging?

I tried to make this possible but the Django config does not allow the variable to be optional. Maybe someone more familiar with Django can figure this out. It should be possible to use a fake key like `01234567-89ab-cdef-ghij-klmnopqrstuv`.

> should we setup different keys for staging and prod?

No need, we can filter the logs on the dashboard

szabozoltan69:
Now there is some issue on staging... (502 bad gateway)

szabozoltan69:
Now it is fine.

thenav56:
@batpad All looks good to me.

@gulfaraz For defining an optional variable. You can do something like this.
```python
APPLICATION_INSIGHTS_INSTRUMENTATION_KEY = os.environ.get('APPLICATION_INSIGHTS_INSTRUMENTATION_KEY', 'this-is-a-random-key-for-local-devlo')
```
Could you also let us know what happens when the wrong key is provided (or it is run on local)? From the look of it, the invalid key is silently ignored.
Also, how does the client send the data to the server? Is there any authentication? Asking since we have shared the key publicly.

PR #1075, opened 2021-05-19
set active personal based on hidden or draft

geohacker:
Based on what @batpad's instruction. @szabozoltan69 could you test this and deploy to staging? Thank you.

cc @tovari


vdeak:
Deployed to Staging.

PR #1077, opened 2021-05-19
Feature/molnix hidden

batpad:
Addresses failing sync_molnix task due to associated Event not having a country associated.

We country require Personnel Deployments to have countries associated, so currently it will warn of this error in the logs, but continue importing rest of the deployments.

cc @vdeak @tovari


vdeak:
Deployed to Staging

PR #1079, opened 2021-05-21
trim whitespace from ns names coming from molnix api, fix warning for…

batpad:
Couple of minor fixes for the Molnix data import:

 - Trim whitespace from national society names coming from Molnix API
 - Log warning about events without countries correctly to cronjob log in the admin

@vdeak would be great if you could deploy this to staging at some point. thank you!


vdeak:
Deployed to Staging

PR #1080, opened 2021-05-22
Feature/erp tests

szabozoltan69:
Here we do not use API endpoints, so no snapshot test could be used. But unit test and integration test (from API) is done.

### DIFF
--- a/api/test_models.py
+++ b/api/test_models.py
@@ -1,12 +1,10 @@
 from django.test import TestCase
 from django.contrib.auth.models import User
-from rest_framework.test import APIRequestFactory
 from rest_framework.test import APITestCase
 from main.mock import erp_request_side_effect_mock
 from unittest.mock import patch
 
 import api.models as models
-import api.drf_views as views
 from api.factories import country as countryFactory
 from api.factories import event as eventFactory
 from api.factories import field_report as fieldReportFactory
@@ -107,8 +105,12 @@ class FieldReportTest(TestCase):
         country = countryFactory.CountryFactory()
         event = eventFactory.EventFactory(name='disaster2', summary='test disaster 2', dtype=dtype)
         event.countries.set([country])
-        report = fieldReportFactory.FieldReportFactory.create(rid='test2', dtype=dtype, ns_request_assistance=True)
-        # report = models.FieldReport.objects.create(rid='test2', event=event, dtype=dtype, ns_request_assistance=True)
+        report = fieldReportFactory.FieldReportFactory.create(
+            rid='test2',
+            event=event,
+            dtype=dtype,
+            ns_request_assistance=True,
+        )
         ERP = models.ERPGUID.objects.get(api_guid='FindThisGUID')
         self.assertEqual(ERP.field_report_id, report.id)
         self.assertEqual(mocked_requests_post.called, True)
--- a/api/test_models.py
+++ b/api/test_models.py
@@ -107,12 +107,11 @@ class FieldReportTest(TestCase):
         country = countryFactory.CountryFactory()
         event = eventFactory.EventFactory(name='disaster2', summary='test disaster 2', dtype=dtype)
         event.countries.set([country])
-        report = fieldReportFactory.FieldReportFactory.create(rid='test2', dtype=dtype, ns_request_assistance=True) # Didn't work via factory, it didn't call push_fr_data
-        report.save()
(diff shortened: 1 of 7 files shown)


utils/test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+                self.json_data = json_data
+                self.text       = 'FindThisGUID'
+                self.status_code = status_code
+
+            def json(self):
+                return self.json_data
+
+        if not kwargs['json']['Emergency']['FieldReport']['AffectedCountries']:
+            return MockResponse(None, 400)
+        return MockResponse(None, 200)
```
thenav56, line 20: Let's define this outside the ERPTest class. This way we can reuse this in `api/test_models.py` as well (keeping a single mock source).

Also, Instead of MockResponse class, we can use `Mock` here. Something like this.
```python3
from unittest.mock import Mock
....

def erp_request_side_effect_mock(url, json):
    def _generate_mock(status_code, json):
        response_mock = Mock()
        response_mock.text = 'FineThisGUID'
        response_mock.status_code = status_code
        response_mock.json.return_value = json
        return response_mock

    if json['Emergency']['FieldReport']['AffectedCountries']:
        return _generate_mock(200, None)
    return _generate_mock(400, None)
```

api/test_models.py
```diff
@@ -96,6 +97,29 @@ def test_field_report_create(self):
...
+    @mock.patch('utils.erp.requests.post', side_effect=mocked_requests_post)
+    def test_ERP_related_field_report(self, mocked_requests_post):
+        dtype = models.DisasterType.objects.get(pk=1)
+        country = models.Country.objects.create(name='country', iso='CH')
+        event = models.Event.objects.create(name='disaster2', summary='test disaster 2', dtype=dtype)
+        event.countries.set([country])
+        report = models.FieldReport.objects.create(rid='test2', event=event, dtype=dtype, ns_request_assistance=True)
+        self.assertEqual(mocked_requests_post.called, True)
+        ERP = models.ERPGUID.objects.get(api_guid='FindThisGUID')
+        self.assertEqual(ERP.field_report_id, report.id)
```
thenav56, line 121: For `Model.objects.create(` let's use factory models instead.

thenav56 (changes requested):

utils/test_utils.py
szabozoltan69: Done.

api/test_models.py
szabozoltan69: I created a FieldReport factory (it did not exist before, in api), but via factory the field report creation (ns_required_assistance=True) did not call the needed method. So I created the other entities via factory, but the field report is created in the old fashioned way.
Nevertheless it was a good practice to crate a new (complicated) factory.

thenav56: This statement will only create a object in memory `fieldReportFactory.FieldReportFactory(rid='test2', dtype=dtype, ns_request_assistance=True)`. You need to use `FieldReportFactory.create(..)` or save report instance to trigger `push_fr_data`.

szabozoltan69: Well, I could not achieve to trigger push_fr_data via the factory way.
It always gave:
api.models.ERPGUID.DoesNotExist: ERPGUID matching query does not exist.
(and if I debug it, models.ERPGUID.objects.all() is really empty.)

If you can fix it, I would greatly appreciate the help.

szabozoltan69: (Hmm... the local test differs from the CircleCI one. Locally self.assertEqual(erp_request_side_effect_mock.called, False) worked, but remotely it failed.)

thenav56: You forgot to pass the event.
I encounter another issue with the factory, when `post_generation` is used, the factory will trigger the model save() **twice** as described [here](https://github.com/FactoryBoy/factory_boy/issues/316). This results in creating **2 ERPGUID**.

If we fix this `TODO: maybe add a check, and only send request if anything has changed`, then it should work as expected.

NOTE: My changes are [here](https://github.com/IFRCGo/go-api/pull/1080/commits/e5586ea765fe95aed2266c74f1853e2cfe1bbffd).

docker-compose exec serve bash -c 'pytest --reuse-db api/test_models.py::FieldReportTest::test_ERP_related_field_report'

szabozoltan69: I could solve this via commenting out the @factory.post_generation parts in field_report factory.
(The check for double call and allowing only the first seemed to be not a really clear solution.)

In test_utils also used factories.

szabozoltan69:
Is it OK like this now @thenav56 ?

thenav56 (approved):

PR #1083, opened 2021-05-26
time machine fixes 1

vdeak:


szabozoltan69:
Hi Viktor! The situation is: in develop branch the api/management/commands/triggers_to_db.py is CRLF terminated. That is why it seems to be conflicting.
In case of merging just use your new (LF terminated) file, it will work fine.

(I shall check how many CRLF terminated file exists in develop now. Should be fixed all.)

szabozoltan69:
(Recently they are CRLF terminated:
deployments/drf_views.py
api/management/commands/triggers_to_db.py
api/views.py
api/drf_views.py
api/serializers.py

I found them via: find . -not -type d -exec file "{}" ";" | grep CRLF

api/management/commands/triggers_to_db.py
```diff
@@ -73,5 +73,54 @@ def handle(self, *args, **options):
...
+       OR (OLD.amount_requested  IS DISTINCT FROM NEW.amount_requested)
+       OR (OLD.amount_funded     IS DISTINCT FROM NEW.amount_funded))
+    EXECUTE PROCEDURE log_api_appeal_changes();
+
+
+DROP TRIGGER IF EXISTS ti_api_appeal_01 on api_appeal;
+CREATE TRIGGER ti_api_appeal_01 BEFORE INSERT
+ON api_appeal FOR EACH ROW
+    EXECUTE PROCEDURE log_api_appeal_insterts();
+
```
thenav56, line 123: Let's use the Django signal for this. (in `api/receivers.py`)

Here is a sample.
```python3
@receiver(post_save, sender=Appeal)
def add_update_appeal_history(sender, instance, created, **kwargs):
    fields_watched = [
        field.name for field in AppealHistory._meta.get_fields()
        if field.name not in ['id', 'appeal', 'valid_from', 'valid_to']
    ]
    now = timezone.now()
    latest_attributes = {
        field: getattr(instance, field)
        for field in fields_watched
    }.update(dict(
        appeal=instance,
        valid_from=now,
        # valid_to=  # Let's leave this to null and when we filter in the views.py lets coalesce this with now date
    ))

    if created:  # Appeal Insert
        AppealHistory.objects.create(**latest_attributes)
    else:  # Appeal Update
        appeal_history = AppealHistory.objects.filter().order_by('id').last()
        if appeal_history and not any([getattr(instance, field) != getattr() for field in fields_watched]):
            # Watched fields are not changed
            return
        appeal_history.valid_to = now
        appeal_history.save(update_fields='valid_to')
        # Create a new appeal history if changed
         AppealHistory.objects.create(**latest_attributes)
```

api/models.py
```diff
@@ -893,6 +893,32 @@ def __str__(self):
...
+    country = models.ForeignKey(Country, verbose_name=_('country'), null=True, on_delete=models.SET_NULL)
+    region = models.ForeignKey(Region, verbose_name=_('region'), null=True, on_delete=models.SET_NULL)
+
+    class Meta:
+        ordering = ('-start_date', '-end_date',)
+        verbose_name = _('appealhistory')
+        verbose_name_plural = _('appealhistories')
+
+    def record_type(self):
+        return 'APPEALHISTORY'
```
thenav56, line 918: Is this used? If yes let's use an enum instead of a manual string.

thenav56 (changes requested):

PR #1084, opened 2021-05-27
deal with change in molnix api returning secondment incoming ns name

batpad:
Deals with a change in the Molnix API -

Secondment Incoming NS is now included within an `incoming` object. Corrects the parser to use the new field.


batpad:
@vdeak possible to merge and deploy this to staging whenever possible? Thanks!

PR #1105, opened 2021-06-11
handle case where incoming value for personnel deployment from molnix is null

batpad:
It is possible for the Molnix API to return null values for the secondment_incoming for a deployment. We were not handling this case and always expecting a value.

This fixes that and adds a message to the log output with the Molnix ID of the entry that did not have `incoming` data.

cc @vdeak @szabozoltan69 - whenever possible, if you could review, merge and deploy to staging, would be great (no rush). Thanks.


szabozoltan69 (approved):

szabozoltan69:
@batpad – Merged to develop and deployed to staging

PR #1112, opened 2021-06-22
update geom if it exists instead of adding new

geohacker:
Address an issue in the import script to not create new geometries. @batpad could you review?

### DIFF
--- a/api/management/commands/import-admin0-data.py
+++ b/api/management/commands/import-admin0-data.py
@@ -117,11 +117,17 @@ class Command(BaseCommand):
           country = Country.objects.get(iso=feature_iso2, record_type=1)
 
           if options['update_geom']:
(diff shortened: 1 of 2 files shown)


batpad:
Ah, I see, you were correctly updating the district / country, but creating a new row for the geometry - ha, this looks good.

@geohacker should it be safe to just merge this and re-run the import command, or would we need to do anything else in the db before running the command?

This fix looks good to me.

batpad (approved):

geohacker:
We should be able to rerun, might have to manually remove the duplicate geom on staging but need to confirm the import actually happened.

szabozoltan69:
Deployed to staging, now re-running the two (different) import-admin0-data commands...

PR #1114, opened 2021-06-23
Over-ride matching for NS names from Molnix

batpad:
Refs https://github.com/IFRCGo/go-frontend/issues/1638#issuecomment-859368247

For certain NS names from Molnix, specify a manual matching to Country entities in GO.

cc @szabozoltan69 @vdeak - please merge and deploy to staging whenever able.

### DIFF
--- a/api/management/commands/sync_molnix.py
+++ b/api/management/commands/sync_molnix.py
@@ -12,6 +12,23 @@ from api.create_cron import create_cron_record
 
 CRON_NAME = 'sync_molnix'
 
+'''
+    Some NS names coming from Molnix are mapped to "countries"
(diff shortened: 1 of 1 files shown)


szabozoltan69 (approved):
If these overrides change frequently, could be kept in database. But for now it is fine like this.

batpad:
> If these overrides change frequently, could be kept in database. But for now it is fine like this.

Yes. Ideally in the future we would just have these entities on GO and not need these over-rides, but yes, if we do, it maybe nice to consider moving this to the db.

szabozoltan69:
Deployed to staging

PR #1115, opened 2021-06-23
Hotfix/update geo imports

batpad:
This contains just the commits to fix the geo-imports straight to `master`, so that we are able to run these on production without including all changes from `develop`.

cc @geohacker @szabozoltan69


szabozoltan69:
For me it is ok.
(A quiestion: do we want to run only the import-admin0/1... commands, or also update-mapbox on production?)

batpad:
@szabozoltan69 I believe we want to do both -- first, run the import-admin0/1, and then run the update-mapbox on production (which would need to be done from the new VM) -- if it makes it easier, I think both the steps can be done from the new VM.

PR #1116, opened 2021-06-24
filter out non-independent countries when matching NS names

batpad:
Countries data on staging has been updated so that there can now be multiple "Country" entities in the db with the same National Society name. This was causing the ingestion to fail as we need to match to a single country.

This adds a check for `independent=True` as we never want to associate Deployments with non-Independent countries - they should always be associated with the "main" independent=True country entity.

cc @vdeak @szabozoltan69 - sorry for the constant small bugfixes on this - these have been hard to reproduce when developing locally. Please let me know if you're able to deploy this to staging sometime. Thank you!!


szabozoltan69 (approved):

szabozoltan69:
Will be on staging in 10 minutes...

PR #1117, opened 2021-06-29
Ignore instrumentation key if not provided

thenav56:
Addresses CI/CD failure in master.

## Changes

* Ignore instrumentation key if not provided.


szabozoltan69 (approved):
Great find and change.

PR #1120, opened 2021-06-30
Feature/covid page related updates

thenav56:
Addresses https://github.com/IFRCGo/go-frontend/issues/1950 

## Changes
- Add EventFeaturedDocument, EventLink inline to event.
- Add EventFeaturedDocument, EventLink to translation (title, description).
- Add EventFeaturedDocument, EventLink to Event Detail API.
    - Add snapshot test for Event Detail.
- Fix flake8 issues.
- Add `/media/` route for DEBUG Mode for serving media files using the local disk.


szabozoltan69 (approved):

szabozoltan69:
[Hi Navin, now I am on leave until mid-August.]

thenav56:
> [Hi Navin, now I am on leave until mid-August.]

Sorry for disturbing @szabozoltan69.

@vdeak Could you review and deploy to staging if all is good?

api/models.py
```diff
@@ -632,6 +644,43 @@ def __str__(self):
...
+        verbose_name=_('event'),
+        on_delete=models.CASCADE,
+        related_name="featured_documents",
+        related_query_name="featured_document",
+    )
+    title = models.CharField(verbose_name=_('title'), max_length=200)
+    description = models.TextField(verbose_name=_('description'))
+    thumbnail = models.ImageField(
+        verbose_name=_('thumbnail'), upload_to='event-featured-documents/thumbnail/',
+        help_text=_('Please maintain aspect ratio (3:4) of image while uploading'),
```
batpad, line 659: minor: is that aspect ratio right? a more standard aspect ratio is 4:3 - should we just clarify width vs height to make this clearer?

batpad:
@thenav56 this looks spectacular to me - thank you for all the lintfixes and tests. Left a very minor note / question around image aspect ratio.

Moving the datetime mocks one level up on the tests looks great -- thank you for all this.

My only concern here: this is going to be very hard to untangle from develop if we want to hold this change back for this deploy. It seems to me like even if the frontend approval of this takes some time, going ahead with merging this into the backend shouldn't break anything - @thenav56 can you confirm, if it seems safe to merge this in, and deploy to prod, even if, for any reason, we need to hold back the frontend changes here?

thenav56:
> @thenav56 this looks spectacular to me - thank you for all the lintfixes and tests. Left a very minor note / question around image aspect ratio.
> 
> Moving the datetime mocks one level up on the tests looks great -- thank you for all this.
> 
> My only concern here: this is going to be very hard to untangle from develop if we want to hold this change back for this deploy. It seems to me like even if the frontend approval of this takes some time, going ahead with merging this into the backend shouldn't break anything - @thenav56 can you confirm, if it seems safe to merge this in, and deploy to prod, even if, for any reason, we need to hold back the frontend changes here?

Thanks for the review. I think it's safe to merge before merging the frontend.

api/models.py
thenav56: @frozenhelium is suggesting `96x128`. Is this okay?

batpad: If we do want something like 96x128, then 3:4 should be fine -- are we enforcing some sort of max upload size or so? I'm guessing we are not doing any thumbnailing ourselves so relying on the user to resize images here? Then it might make sense to give the suggestion in pixel values + maybe enforce a max size somewhere? Users will inevitably try and upload a 30MB image file and we should be prepared :-)

batpad (approved):

api/models.py
frozenhelium: Since the image is in portrait format the aspect ratio would be 3:4
@thenav56 may be we can make following the help text?
`For the better result please make sure the thumbnail has resolution close to 96x128`
Does that look good @batpad @tovari

batpad: To make this a bit clearer for users, can we say:

`Image should be portrait (3:4 aspect ratio) and scaled down to as close to 96x128 as the image size`

@thenav56 if not for the current PR, can we ticket enforcing some sort of sane max image size on uploads?

batpad:
> Thanks for the review. I think it's safe to merge before merging the frontend.

Yep, this seems safe to me - @thenav56 go ahead and merge this to develop once the help_text for images is resolved, and @vdeak after that, let us know when you're able to deploy to staging so folks can start testing the frontend as soon as possible. Thanks!

api/models.py
thenav56: Update the help text and create a new issue for adding validation https://github.com/IFRCGo/go-api/issues/1121.

batpad:
@vdeak please deploy latest `develop` to staging whenever possible. Thank you.

vdeak:
Deployed to Staging

Issue #1092, opened 2021-06-02
3W: Global view: Create Aggregated API


thenav56:
Create/Update API endpoint according to this mockup
https://www.figma.com/proto/RZTZdDWf6Xjweh9GA8rMGW/IFRC_GO_Wires_20210502?page-id=0%3A72&node-id=111%3A0&viewport=1350%2C3308%2C0.12138918787240982&scaling=scale-down-width&hide-ui=1

Related to https://github.com/IFRCGo/go-frontend/issues/1823

thenav56:
This is deployed to staging. It can be tested from go-frontend https://github.com/IFRCGo/go-frontend/issues/1823 so closing this issue.

PR #1124, opened 2021-07-02
Use ongoing projects for global project Endpoint

thenav56:
Addresses  3W: Global view: Create Aggregated API #1092  
cc: @frozenhelium 

## Changes

* Only use ongoing projects.

### DIFF
--- a/deployments/drf_views.py
+++ b/deployments/drf_views.py
@@ -406,7 +406,13 @@ class GlobalProjectViewset(ReadOnlyVisibilityViewsetMixin, viewsets.ViewSet):
         # Filter by visibility
         qs = self.get_visibility_queryset(Project.objects.all())
         # Filter by GET params
-        return ProjectFilter(self.request.query_params, queryset=qs).qs
+        projects = ProjectFilter(self.request.query_params, queryset=qs).qs
+        return Project.objects.filter(
+            # To avoid duplicate rows
+            id__in=projects,
+            # NOTE: Only process ongoing projects in global project view
+            status=Statuses.ONGOING,
+        )
 
     @action(detail=False, url_path='overview', methods=('get',))
     def overview(self, request, pk=None):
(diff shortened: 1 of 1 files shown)


vdeak (approved):

deployments/drf_views.py
```diff
@@ -406,7 +406,13 @@ def get_projects(self):
         # Filter by visibility
         qs = self.get_visibility_queryset(Project.objects.all())
         # Filter by GET params
-        return ProjectFilter(self.request.query_params, queryset=qs).qs
+        projects = ProjectFilter(self.request.query_params, queryset=qs).qs
+        return Project.objects.filter(
```
batpad, line 410: I'm sure there's a good reason, but why not just `project.objects.filter(...)` here and why do we need the `id__in=projects` here? Just curious.

batpad (approved):

deployments/drf_views.py
thenav56: I am not sure, but I remember getting duplicate rows and wasn't able to use distinct() here. So to be sure, I am passing the filtered queryset as a subquery.

batpad: I see, could we add a comment to that effect?

Strange that distinct doesn't work, but I know I've had the problem where it's some-how really hard to get rid of duplicates with complex joins - have you tried something like `.distinct('id')` or so? Anyways, this looks fine - might just be nice to have a comment.

thenav56: Sorry, but there is a comment. Do you want me to extend it?

batpad: Sorry, my bad - no, this is fine. Shall I go ahead and merge this?

thenav56: Yes please merge and deploy to staging.

frozenhelium:
@vdeak Can you please deploy this to staging?

vdeak:
@frozenhelium It is deployed to staging

frozenhelium:
Thanks :)

PR #1133, opened 2021-07-14
management command to update/create geoms from a shapefile with ids

geohacker:
This is to update some existing country geometries that happen to have the wrong projection. We'll do this as a one off update with ids in the shapefile. cc @batpad @jhenshall


geohacker:
I've not tested this locally just yet. Will wait for a sample shapefile fro m @jhenshall.

batpad (approved):

PR #1140, opened 2021-07-26
Fix country association with Molnix imports

batpad:
Should fix associating a deployment with IFRC: don't check for independent flag when matching by country name.

@vdeak could you merge to develop / deploy whenever possible? (no rush)


vdeak:
@batpad It is deployed to Staging

PR #1141, opened 2021-07-29
Model setup for dref module

k9845:
Addresses https://github.com/IFRCGo/go-frontend/issues/1948 

## Changes
- Created separate app for `dref`
## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


dref/admin.py
```diff
@@ -0,0 +1,48 @@
...
+    extra = 0
+
+
+@admin.register(Dref)
+class DrefAdmin(TranslationAdmin, admin.ModelAdmin):
+    search_fields = ('title',)
+    list_display = ('title', 'national_society', 'disaster_type',
+                    'ns_request_date', 'submission_to_geneva', 'status')
+    inlines = [DrefCountryDistrictAdminInline]
+
```
thenav56, line 45: Can you check if we need autocomplete fields here?

dref/enums.py
```diff
@@ -0,0 +1,91 @@
+import enum
```
thenav56, line 1: Let's add a small note telling why we are defining this here, what is the source and when we can remove this (after upgrading Django to 3)

dref/models.py
```diff
@@ -0,0 +1,568 @@
...
+            ORANGE = _('Orange')
+            RED = _('Red')
+
+    class Status(IntEnum):
+        IN_PROGESS = 0
+        COMPLETED = 1
+
+        class Labels:
+            IN_PROGESS = _('In Progress')
+            COMPLETED = _('Completed')
```
thenav56, line 184: Can we switch to IntegerChoices here?

dref/models.py
```diff
@@ -0,0 +1,568 @@
...
+    class Meta:
+        verbose_name = _('dref')
+        verbose_name_plural = _('drefs')
+
+    def save(self, *args, **kwargs):
+        # TODO: Need to remove this if not required later
+        if self.date_of_approval:
+            self.status = Dref.Status.COMPLETED
+        elif not self.date_of_approval:
+            self.status = Dref.Status.IN_PROGESS
```
thenav56, line 539: Is this still required?

Also we can use single conditional here.
```suggestion
        self.status = Dref.Status.COMPLETED if self.date_of_approval else Dref.Status.IN_PROGESS
```

dref/models.py
```diff
@@ -0,0 +1,568 @@
...
+    district = models.ManyToManyField(District, blank=True,
+                                      verbose_name=_('district'))
+
+    class Meta:
+        unique_together = ('dref', 'country')
+
+
+class DrefFile(models.Model):
+    file = models.FileField(
+        blank=True, null=True, verbose_name=_('file'),
```
thenav56, line 558: Is `blank=True, null=True` needed? DrefFile main responsibility is to attach a file right?

dref/scripts.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def pdf_image_extractor(file):
+    pages = convert_from_path(file)
+    for i in range(len(pages)):
+        page = pages[i]
+        page.save(f'{file}_preview.png', 'PNG')
+        with open(f'{file}_preview.png') as f:
+            _, tail = os.path.split(f.name)
+            return os.path.join('dref/images', tail)
```
thenav56, line 13: Not sure how this works?

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class DrefSerializer(
+    EnumSupportSerializerMixin,
+    NestedUpdateMixin,
+    NestedCreateMixin,
+    ModelSerializer
+):
+    MAX_NUMBER_OF_IMAGES = 6
+    ALLOWED_BUDGET_FILE_EXTENSIONS = ["pdf"]
```
thenav56, line 128:
```suggestion
    ALLOWED_BUDGET_FILE_EXTENSIONS = ["pdf"]
    
```

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+        disability_people_per = data.get('disability_people_per', '')
+        if disability_people_per and len(disability_people_per.split('.')[0]) == 3:
+            data['disability_people_per'] = disability_people_per.split('.')[0]
+        people_per_urban = data.get('people_per_urban', '')
+        people_per_local = data.get('people_per_local', '')
+        if people_per_urban and len(people_per_urban.split('.')[0]) == 3:
+            data['people_per_urban'] = people_per_urban.split('.')[0]
+        if people_per_local and len(people_per_local.split('.')[0]) == 3:
+            data['people_per_local'] = people_per_local.split('.')[0]
+        return data
```
thenav56, line 162: Why are we doing this?

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def validate(self, data):
+        event_date = data.get('event_date', None)
```
thenav56, line 165:
```suggestion
        event_date = data.get('event_date')
```

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+        if event_date and data['type_of_onset'] not in [Dref.OnsetType.SLOW, Dref.OnsetType.SUDDEN]:
+            raise serializers.ValidationError({
+                'event_date': ugettext('Cannot add event_date if onset type not in {} or {}')
+                .format(Dref.OnsetType.SLOW.label, Dref.OnsetType.SUDDEN.label)
```
thenav56, line 169: Is `.format` compatible with ugettext?

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+                for image in (self.instance and self.instance.images.all()) or []
+            }
+            for image in images:
+                # Ignore for current images (if file is not changed)
+                if image.id is not None and image.id in current_images_map and image.file == current_images_map[image.id]:
+                    continue
+                elif image.created_by != self.context['request'].user:
+                    raise serializers.ValidationError(
+                        ugettext('Only image owner can attach image(id: %s)' % image.id)
+                    )
```
thenav56, line 188: - Can we update DrefFile? I would recommend removing this action. Basically, each DrefFile represents a file, so only create/delete action makes sense.
- So, if we can't update DrefFile, then we can just do something like this to check if there are any images to which the user doesn't have access. `user .values_list('id', flat=True)` if we need to send id back as well)
```python3
images_id = [image.id for image in images]
Dref.objects.filter(
    # Not created by current user
    ~models.Q(created_by=self.context['request']),
    # Look into provided images
    id__in=images_id
).exclude(
    # Exclude already attached images
    id__in=self.instance.images.all()
).exists()
```

dref/serializers.py
```diff
@@ -0,0 +1,217 @@
...
+        # Don't allow images more than MAX_NUMBER_OF_IMAGES
+        if len(images) > self.MAX_NUMBER_OF_IMAGES:
+            raise serializers.ValidationError(
+                ugettext('Can add utmost %s images' % self.MAX_NUMBER_OF_IMAGES)
+            )
+        return images
+
+    def validate_extension(self, filename):
```
thenav56, line 196: Is this used by other than validate_budget_file or by extension field? If not let's scope this inside validate_budget_file

dref/views.py
```diff
@@ -0,0 +1,152 @@
...
+    PlannedInterventionSerializer,
+    IdentifiedNeedSerializer,
+    DrefFileSerializer,
+    IdentifiedNeedSerializer,
+)
+from dref.filter_set import DrefFilter
+
+
+class DrefViewSet(viewsets.ModelViewSet):
+    # TODO: Need to add permission to delete?
```
thenav56, line 32: Is this done?

dref/views.py
```diff
@@ -0,0 +1,152 @@
...
+class PlannedInterventionViewSet(viewsets.ModelViewSet):
+    serializer_class = PlannedInterventionSerializer
+    permission_classes = [permissions.IsAuthenticated]
+    queryset = PlannedIntervention.objects.all()
+
+
+class IdentifiedNeedViewSet(viewsets.ModelViewSet):
+    serializer_class = IdentifiedNeedSerializer
+    permission_class = [permissions.IsAuthenticated]
+    queryset = IdentifiedNeed.objects.all()
```
thenav56, line 123: Don't we need user-level permissions here?

dref/views.py
```diff
@@ -0,0 +1,152 @@
...
+    def multiple_file(self, request, pk=None, version=None):
+        # converts querydict to original dict
+        files = dict((request.data).lists())['file']
+        data = [{'file': file} for file in files]
+        file_serializer = DrefFileSerializer(data=data, context={'request': request}, many=True)
+        if file_serializer.is_valid():
+            file_serializer.save()
+            return response.Response(file_serializer.data, status=status.HTTP_201_CREATED)
+        else:
+            return response.Response(file_serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
thenav56, line 152: Let's save after all are validated. If we don't do that then files will be uploaded but the DrefFile will be rolled back creating orphan files in the server.

main/writable_nested_serializers.py
```diff
@@ -0,0 +1,425 @@
+# -*- coding: utf-8 -*-
```
thenav56, line 1: Is this copied from deep or drf_writable_nested?
if it's drf_writable_nested then we can just add this to requirement.txt else fine.

thenav56 (changes requested):

dref/views.py
thenav56: Mistake review.

main/writable_nested_serializers.py
k9845: Copied from `deep`

dref/views.py
k9845: Basically we don't need these, removed them

dref/views.py
k9845: anyone can delete, who have access to dref.

dref/serializers.py
k9845: moved

dref/scripts.py
k9845: we are generating image from a pdf file uploaded, the image is the first page of pdf

dref/serializers.py
```diff
@@ -0,0 +1,220 @@
...
+        for image in not_created_by_user_images:
+            raise serializers.ValidationError(
+                ugettext('Only image owner can attach image(id: %s)' % image)
+            )
+        # Don't allow images more than MAX_NUMBER_OF_IMAGES
+        if len(images) > self.MAX_NUMBER_OF_IMAGES:
+            raise serializers.ValidationError(
+                ugettext('Can add utmost %s images' % self.MAX_NUMBER_OF_IMAGES)
+            )
+        return images
```
thenav56, line 200: Continuation of https://github.com/IFRCGo/go-api/pull/1141#discussion_r798229277

```suggestion
        # Don't allow images more than MAX_NUMBER_OF_IMAGES
        if len(images) > self.MAX_NUMBER_OF_IMAGES:
            raise serializers.ValidationError(
                ugettext('Can add utmost %s images' % self.MAX_NUMBER_OF_IMAGES)
            )
        images_id = [image.id for image in images]
        images_without_access_qs = DrefFile.objects.filter(
            # Not created by current user
            ~models.Q(created_by=self.context['request'].user),
            # Look into provided images
            id__in=images_id,
        )
        # Exclude already attached images if exists
        if self.instance:
            images_without_access_qs = images_without_access_qs.exclude(
                id__in=self.instance.images.all()
            )
        images_id_without_access = images_without_access_qs.values_list('id', flat=True)
        if images_id_without_access:
            raise serializers.ValidationError(
                ugettext(
                    'Only image owner can attach image. Not allowed image ids: %s' % ','.join(map(str, images_id_without_access))
                )
            )
        return images
```

dref/serializers.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        disability_people_per = data.get('disability_people_per', '')
+        people_per_urban = data.get('people_per_urban', '')
+        people_per_local = data.get('people_per_local', '')
+        data['disability_people_per'] = self.remove_digits_after_decimal(disability_people_per)
+        data['people_per_urban'] = self.remove_digits_after_decimal(people_per_urban)
+        data['people_per_local'] = self.remove_digits_after_decimal(people_per_local)
+        return data
```
thenav56, line 168: Continuation of https://github.com/IFRCGo/go-api/pull/1141#discussion_r798225677

```suggestion
    def to_representation(self, instance):
        def _remove_digits_after_decimal(data, key):
            # NOTE: We are doing this to remove decimal after 3 digits whole numbers
            # eg: 100.00% be replaced with 100%
            if value and len(value.split('.')[0]) == 3:
                return value.split('.')[0]
            return None

        data = super().to_representation(instance)
        for key in [
            'disability_people_per',
            'people_per_urban',
            'people_per_local',
        ]:
            value = data.get(key) or ''
            data[key] = _remove_digits_after_decimal(value)
        return data
```

dref/serializers.py
```diff
@@ -0,0 +1,220 @@
...
+        return images
+
+    def validate_budget_file(self, budget_file):
+        if budget_file:
+            extension = os.path.splitext(budget_file.file.name)[1].replace(".", "")
+            if extension.lower() not in self.ALLOWED_BUDGET_FILE_EXTENSIONS:
+                raise serializers.ValidationError(
+                    f'Invalid uploaded file extension: {extension}, Supported only PDF Files'
+                )
+        return budget_file
```
thenav56, line 209:
```suggestion
        if budget_file is None:
            return
        extension = os.path.splitext(budget_file.file.name)[1].replace(".", "")
        if extension.lower() not in self.ALLOWED_BUDGET_FILE_EXTENSIONS:
            raise serializers.ValidationError(
                f'Invalid uploaded file extension: {extension}, Supported only PDF Files'
            )
        return budget_file
```

thenav56 (changes requested):

dref/scripts.py
thenav56: I get that, but how are we storing the file? Using file storage? So, how the `page.save()` and `dref/images` works here? Will this work when we have azure/s3 storage?

dref/views.py
thenav56: okay

dref/views.py
thenav56: haha

dref/scripts.py
k9845: we are using file storage, `page.save()` will create a new image from pdf, we use that new file and store that in our database.

thenav56: Will this work when we have azure/s3 storage?

k9845: yes, since this works like we upload file as in other file field

dref/models.py
```diff
@@ -518,6 +517,16 @@ class Meta:
         verbose_name_plural = _('drefs')
 
     def save(self, *args, **kwargs):
+        if self.budget_file:
```
thenav56, line 520: We also need to check if this has changed. Else this will be triggered for every save.

dref/models.py
```diff
@@ -518,6 +517,16 @@ class Meta:
...
+            pages = convert_from_path(self.budget_file.file.path)
+            for i in range(len(pages)):
```
thenav56, line 522: Why for loop?

dref/models.py
```diff
@@ -518,6 +517,16 @@ class Meta:
...
+                budget_file_preview = pages[i]
+                filename = f'preview_{self.budget_file.file.name.split("/")[0]}.png'
+                temp_image = open(os.path.join('/tmp', filename), 'wb')
+                budget_file_preview.save(temp_image, 'PNG')
+                thumb_data = open(os.path.join('/tmp', filename), 'rb')
+                self.budget_file_preview.save(filename, thumb_data, save=False)
```
thenav56, line 528: Why change the same field when in a loop?

thenav56 (changes requested):

thenav56 (approved):

frozenhelium:
@szabozoltan69 this PR is also ready to go to the staging server. Please merge and deploy at your convenience.

szabozoltan69:
The local migration was not successful, so I must find out the reason and smooth it (on database level), because I receive this:
```
Running migrations:
  Applying api.0133_auto_20210729_1609...Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: could not create unique index "api_appealhistory_code_6daba631_uniq"
DETAIL:  Key (code)=(MDR65001) is duplicated.
```

k9845:
@szabozoltan69 `api.0133_auto_20210729_1609` and `0133_auto_20210922_1327` seems out to be same migrations, may be we hsould unapply one of them....this might cause later merge migrations to be removed as well

szabozoltan69:
Ok, I arrange them...

szabozoltan69:
Will be deployed to staging in 10 minutes...

szabozoltan69:
@frozenhelium @k9845 Deployed to staging

PR #1162, opened 2021-08-23
Should not use XML2Dict==0.2.2 due to security reason. Instead: xmltodict

szabozoltan69:
We used in 2 ingest scripts the XML2Dict – which can be replaced by xmltodict.
https://github.com/IFRCGo/go-api/security/dependabot/requirements.txt/XML2Dict/open
There is no possibility to upgrade, XML2Dict is abandoned: https://pypi.org/project/XML2Dict/#history

ingest_who - is not used nowadays, but
ingest_gdacs is heavily used. Some parts the script had to be changed according to api/management/commands/scrape_pdfs.py

Old structure example: https://pastebin.com/bWrEh0p9
New structure example: https://pastebin.com/YBnWMVWg
Locally tested.

### DIFF
--- a/api/management/commands/ingest_gdacs.py
+++ b/api/management/commands/ingest_gdacs.py
@@ -1,6 +1,6 @@
 import requests
 import datetime as dt
-from encoder import XML2Dict
+import xmltodict
 from dateutil.parser import parse
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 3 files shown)


thenav56 (approved):
Looks good.

PR #1167, opened 2021-08-30
Support multiple file upload

k9845:
Addresses xxxxxx 

## Changes

* ...

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


dref/models.py
```diff
@@ -375,6 +370,17 @@ class Labels:
...
+    event_map = models.ForeignKey(
+        'DrefFile', on_delete=models.SET_NULL,
+        null=True, blank=True,
+        verbose_name=_('event map'),
+        related_name=_('event_map')
+    )
+    images = models.ManyToManyField(
+        'DrefFile', blank=True,
+        verbose_name=_('images'),
+        related_name=_('images')
```
thenav56, line 382: `related_name` is used for reverse relation. Let's use proper names (and remove translation) here (and above).

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey.related_name

dref/models.py
```diff
@@ -401,14 +407,16 @@ class Meta:
...
-    Multiple images are uploaded
-    """
-    dref = models.ForeignKey(Dref, verbose_name=_('dref'),
-                             on_delete=models.CASCADE)
-    image = models.FileField(
-        verbose_name=_('image'), null=True, blank=True,
-        upload_to=image_path, storage=get_storage(),
-        help_text=_('Images of the event')
+def dref_path(instance, filename):
+    return f'dref/{filename}'
```
thenav56, line 411: We don't need to define a function for simple use-cases.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.FileField.upload_to

dref/models.py
```diff
@@ -401,14 +407,16 @@ class Meta:
...
+
+
+class DrefFile(models.Model):
```
thenav56, line 414: Let's also add created_by for ownership. Also, make sure only owners can attach files to DREF.

dref/serializers.py
```diff
@@ -78,7 +77,8 @@ class DrefSerializer(
     disaster_category_level_display = serializers.CharField(source='get_disaster_category_level_display', read_only=True)
     status_display = serializers.CharField(source='get_status_display', read_only=True)
     modified_by_details = UserNameSerializer(source='modified_by', read_only=True)
-    images = DrefImgaeSerializer(source='drefimage_set', many=True, required=False)
+    event_map_details = serializers.CharField(source='event_map', read_only=True)
```
thenav56, line 80: Shouldn't this be DrefFileSerializer?

dref/serializers.py
```diff
@@ -100,6 +100,12 @@ def validate(self, data):
             )
         return data
 
+    def validate_images(self, images):
+        MAX_NUMBER_OF_IMAGES = 6
```
thenav56, line 104: Let's move this to serializer scope.

dref/views.py
```diff
@@ -97,3 +99,38 @@ class IdentifiedNeedViewSet(viewsets.ModelViewSet):
...
+        permission_classes=[permissions.IsAuthenticated],
+    )
+    def multiple_file(self, request, pk=None, version=None):
+        def modify_input_for_multiple_files(file):
+            dict = {}
+            dict['file'] = file
+            return dict
+        # converts querydict to original dict
+        files = dict((request.data).lists())['file']
+        flag = 1
```
thenav56, line 122: Can't we just check `len(arr)`?

dref/views.py
```diff
@@ -97,3 +99,38 @@ class IdentifiedNeedViewSet(viewsets.ModelViewSet):
...
+        arr = []
+        for file in files:
+            modified_data = modify_input_for_multiple_files(file)
+            file_serializer = DrefFileSerializer(data=modified_data)
+            if file_serializer.is_valid():
+                file_serializer.save()
+                arr.append(file_serializer.data)
+            else:
+                flag = 0
```
thenav56, line 131: Can't we do this with many=True?

dref/views.py
```diff
@@ -97,3 +99,38 @@ class IdentifiedNeedViewSet(viewsets.ModelViewSet):
...
+
+        if flag == 1:
+            return response.Response(arr, status=status.HTTP_201_CREATED)
+        else:
+            return response.Response(arr, status=status.HTTP_400_BAD_REQUEST)
```
thenav56, line 136: This willn't work. Since we are `file_serializer.save()`, some files will be saved but the response will be just an error.

thenav56 (changes requested):

dref/serializers.py
```diff
@@ -100,6 +110,15 @@ def validate(self, data):
             )
         return data
 
+    def validate_images(self, images):
+        if images:
+            for image in images:
+                if image.created_by != self.context['request'].user:
+                    raise serializers.ValidationError(ugettext('Can add images that user created'))
```
thenav56, line 117: We also need to check if the image was already attached there. For eg: if user1 owns an image and has attached the image to a DREF, then when another user2 tries to add a new image then this will throw an error.
We need to compare if the file was changed, if not then we can ignore them.

Also, let's change the text.
```suggestion
                    raise serializers.ValidationError(ugettext('Only image owner can attach image id: {image.id}'))
```

dref/serializers.py
```diff
@@ -16,10 +16,13 @@
     NationalSocietyAction,
     IdentifiedNeed,
     DrefCountryDistrict,
-    DrefImage
+    DrefFile
 )
 
 
+MAX_NUMBER_OF_IMAGES = 6
```
thenav56, line 23: Move this to `DrefSerializer`.

Eg:
```
class DrefSerializer(...):
    MAX_NUMBER_OF_IMAGES = 6
```

dref/views.py
```diff
@@ -97,3 +99,28 @@ class IdentifiedNeedViewSet(viewsets.ModelViewSet):
...
+    )
+    def multiple_file(self, request, pk=None, version=None):
+        # converts querydict to original dict
+        files = dict((request.data).lists())['file']
+        arr = []
+        data = [{'file': file} for file in files]
+        file_serializer = DrefFileSerializer(data=data, context={'request': request}, many=True)
+        if file_serializer.is_valid():
+            file_serializer.save()
+            arr.append(file_serializer.data)
```
thenav56, line 123: Can you check if this works?

thenav56 (changes requested):

dref/serializers.py
```diff
@@ -100,6 +108,24 @@ def validate(self, data):
...
+                    if image.id in current_images_id:
+                        if image.file in current_images_file:
+                            continue
+                    elif image.created_by != self.context['request'].user:
+                        raise serializers.ValidationError(ugettext(
+                            'Only image owner can attach image id: {}').format(image.id)
+                        )
+        if len(images) > self.MAX_NUMBER_OF_IMAGES:
+            raise serializers.ValidationError(ugettext('Can add utmost {} images').format(self.MAX_NUMBER_OF_IMAGES))
+        return images
```
thenav56, line 127: Let's do something like this.

```suggestion
    def validate_images(self, images):
        # Validate if images are provided
        if images:
            # Fetch current images
            current_images_map = {
                image.pk: image.file
                for image in (self.instance and self.instance.images.all()) or []
            }
            for image in images:
                # Ignore for current images (if file is not changed)
                if image.id is not None and image.id in current_images_map and image.file == current_images_map[image.id]:
                    continue
                elif image.created_by != self.context['request'].user:
                    raise serializers.ValidationError(
                        ugettext('Only image owner can attach image(id: %s)' % image.id)
                    )
        # Don't allow images more than MAX_NUMBER_OF_IMAGES
        if len(images) > self.MAX_NUMBER_OF_IMAGES:
            raise serializers.ValidationError(
                ugettext('Can add utmost %s images' % self.MAX_NUMBER_OF_IMAGES)
            )
        return images

```

dref/views.py
```diff
@@ -97,3 +99,26 @@ class IdentifiedNeedViewSet(viewsets.ModelViewSet):
     serializer_class = IdentifiedNeedSerializer
     permission_class = [permissions.IsAuthenticated]
     queryset = IdentifiedNeed.objects.all()
+
+
+class DrefFileViewSet(viewsets.ModelViewSet):
```
thenav56, line 104: Let's change this to allow only list, create, and our custom `multiple`. Also, for queryset let's filter by created_at.

Also, let's add some tests as well.

dref/test_views.py
```diff
@@ -209,6 +253,35 @@ def test_event_date_in_dref(self):
...
+        response = self.client.patch(url, data)
+        self.assert_200(response)
+
+        # add from another user
+        data = {
+            "images": [file4.id]
+        }
+        self.client.force_authenticate(self.ifrc_user)
+        response = self.client.patch(url, data)
+        self.assert_200(response)
```
thenav56, line 283: Let's add another for 400 as well.

thenav56 (changes requested):

thenav56 (approved):

PR #1170, opened 2021-08-31
for surge alerts, by default show all, add filter to show only active

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/1965

## Changes

 - Do not filter out `is_active=False` surge alerts by default
 - Add an explicit `is_active` filter to show only active alerts

cc @szabozoltan69


szabozoltan69 (approved):

szabozoltan69:
Deployed to staging

PR #1181, opened 2021-09-08
Fix difference in RR figures and tz

szabozoltan69:
10<->6+3 was wrong.


batpad (approved):

batpad:
Thank you @szabozoltan69 !

PR #1200, opened 2021-09-20
Some quick fixes

szabozoltan69:
## Changes
 - ids are now prefixed with type. add country names etc into district c.
 - Fixing databank source issues (keys)
 - add missing attributes to centroid tilesets
 - Three sleeps into map tile updater
 - Allowing null in region_deployed_to for Molnix


api/management/commands/sync_molnix.py
```diff
@@ -336,6 +336,7 @@ def handle(self, *args, **options):
             countries = molnix.get_countries()
             deployments = molnix.get_deployments()
             open_positions = molnix.get_open_positions()
+            import pdb; pdb.set_trace()
```
batpad, line 339: @szabozoltan69 please remove :)

batpad (approved):
Please remove the pdb.set_trace() before merging.

PR #1210, opened 2021-09-28
Appeal ingest fix

vdeak:


szabozoltan69 (approved):
Seems to be good

vdeak:
How to test on Staging:

1. pick up an appeal which is in the go-api (http://go-api.ifrc.org/api/appeals)
2. modify it manually (set modified_at to years before and modify some values like name, etc...)
3. wait for the Appeal ingest to run (scheduled in every hour)
4. the appeal should be updated (modified_at = date of appeal ingest run; real_data_update = max(APD_modify_time) from the go-api

PR #1212, opened 2021-09-28
Sort the alerts by putting the opens 1st, stood downs 2nd

szabozoltan69:


batpad:
@szabozoltan69 -

So, am fine with merging this right now. But, ideally we should change this to something like:

 - Add another column for `is_stood_down`
 - When ingesting Molnix data, if `status=unfilled`, set `is_stood_down` to `True`
 - Here, order by the `is_stood_down` field instead of the `status` string.

The problem here is if tomorrow Molnix comes up with some new status values or something, it would break the ordering (we are just depending on `unfilled` coming after `open`).

I think we can merge this right now, but could you ticket this change above so we can get to it when there is some time?

Issue #1234, opened 2021-11-19
[PROD] Appeal doc scraper


szabozoltan69:
## Issue
The appeal-document scraper uses 3 months as limit, where it checks the end-date of appeals. The older appeals are not used for document searching.
This interval seems to be to short, some documents appear after 3 months.

We should change it to 6. (This will double the run time of the scraper, but no problem. Instead of 25 appeals it will scrape 50 or so.)

szabozoltan69:
Deployed to staging.
(Also a manual patch on production)

PR #1235, opened 2021-11-19
Appeal doc scraper longer limit

szabozoltan69:


szabozoltan69:
Solves https://github.com/IFRCGo/go-api/issues/1234

PR #1236, opened 2021-11-22
IFRC+NS Content visibility - backend

vdeak:


api/drf_views.py
```diff
@@ -809,8 +817,8 @@ def save_meta(self, fieldreport, meta, is_update=False):
 
 
 class CreateFieldReport(CreateAPIView, GenericFieldReportView):
-    authentication_classes = (TokenAuthentication,)
-    permission_classes = (IsAuthenticated,)
+    #authentication_classes = (TokenAuthentication,)
```
szabozoltan69, line 820: Are you sure that authentication is not required to create a field report?
(I have a fear that robots will create field reports this way.)

deployments/migrations/0042_auto_20211117_1527.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.24 on 2021-11-17 15:27
```
szabozoltan69, line 1: This migration file seems to be removable. Does the same as 0041_auto_20211110_0924.py, except 'IFRC + NS' is 'IFRC_NS', which can be changed manually in 0041_auto_20211110_0924.py.

docker-compose.yml
```diff
@@ -40,7 +40,8 @@ services:
     volumes:
 # Database migration should be done manually, so .tmp -> .db:
       - './.db/pg:/var/lib/postgresql/data'
-
+    ports:
```
szabozoltan69, line 43: Are you sure this ports definition is needed?
(For me everything is fine without it also.)

szabozoltan69 (commented):
3 comments...

szabozoltan69:
Deployed to staging

PR #1277, opened 2022-01-24
Associate Personnel Deployment with country

batpad:
Addresses https://github.com/IFRCGo/go-frontend/issues/2179

Essentially, we used the linked emergency to derive the deployed country for a deployed person. This breaks for multi country emergencies, because obviously, a single person is only deployed to a single country.

This PR fixes this by:

 - Adding a `country_to` column to the individual Personnel model, so individual personnel deployed to the same emergency can have different countries they are deployed to.
 - On the `sync_molnix` cronjob, fetch `country_to` from the `country` field in the deployments API and use that to populate the `country_to` field.

### TODO:

 - Return `country_to` in the Personnel endpoints and use that to display the country a person is deployed to on the frontend.

### Fix Existing Data:  

 - For all existing personnel deployments, we will likely want to populate the `country_to` field using the earlier logic: getting the first country from the associated emergency. @szabozoltan69 will likely need some help here.

@szabozoltan69 sorry, am having some trouble with my local setup - could you pls run `makemigrations` in this branch and commit the migration files for the new column? Then I can roughly add these to the API, but may need your help if my local setup is still broken.

Working on this made me think through a few things about multi-country emergencies in general, will write thoughts separately.

If we can test that ingesting the `country_to` from Molnix is working correctly, we should be able to test this pretty soon - slightly worried about the data migration of old data.

cc @tovari @anamariaescobar


api/management/commands/sync_molnix.py
```diff
@@ -143,8 +143,13 @@ def sync_deployments(molnix_deployments):
...
             p.event_deployed_to = event
 
-            # FIXME: check if country exists, etc.
             if p.event_deployed_to.countries.all().count() > 0:
+                # Since different personnel deployed to the same emergency
+                # can be deployed to different countries affected by the emergency,
+                # we should no longer use country_deployed_to from PersonnelDeployment,
+                # rather get the country for each deployed person directly from the
+                # Personnel model for each deployed person.
+                # FIXME: we should look to deprecate usage of this field entirely.
```
szabozoltan69, line 152: What a  nice long comment!

szabozoltan69:
makemigrations, migrate – DONE

szabozoltan69:
@batpad The database fill-up happens via https://github.com/IFRCGo/go-api/pull/1279/files

szabozoltan69:
Via https://github.com/IFRCGo/go-api/pull/1280/files **country_to** is in the Personnel API endpoints (normal and CSV and anon), pls. try e.g. https://dsgocdnapi.azureedge.net/api/v2/personnel/?limit=4

PR #1294, opened 2022-01-30
Safelink neutralizing

szabozoltan69:


szabozoltan69:
https://stackoverflow.com/questions/62108517/clicking-outlook-safelink-protection-links-in-emails-seems-to-be-executing-the-c

PR #1345, opened 2022-03-07
Registration email check, 2 locations

szabozoltan69:
Changing the limit for the email usage from 1 to 30 days.


batpad:
If we want to try and be more precise we can maybe do something like `months=1` instead of `days=30`, but I also think this, and keeping consistent at 30 days, is fine.

batpad (approved):

PR #1354, opened 2022-03-11
Feature/event name length

vdeak:
Event name length set to 256


szabozoltan69 (approved):

szabozoltan69:
Merged and deployed to staging environment. Migration was run fine.

PR #1356, opened 2022-03-12
Fixing surge deployment non-sending notification

szabozoltan69:


szabozoltan69:
@tovari the mentioned non-sending is fixed via that merge. (At first it is only on staging.)

PR #1360, opened 2022-03-16
add ICRC to matching over-ride list

batpad:
Adds case to correctly match ICRC records during Molnix ingestion.

cc @szabozoltan69


szabozoltan69:
Deployed to staging environment.

PR #1385, opened 2022-03-30
Revision for frontend requests

szabozoltan69:
Using https://django-reversion.readthedocs.io/en/stable/middleware.html as a sample.
(Strange that there is "true", "false" in the code, instead of True, False.)

Maybe some more filtering would be fine (to give which models do we want to save).
This way Field Report changes via frontend are saved – local test:
kép

### DIFF
--- a/middlewares/middlewares.py
+++ b/middlewares/middlewares.py
@@ -4,6 +4,7 @@ from rest_framework import permissions
 from django.conf import settings
 from django.http import JsonResponse
 from django.utils.translation import ugettext, get_language
+# from reversion.middleware import RevisionMiddleware
 
 
 _threadlocal = threading.local()
@@ -67,3 +68,14 @@ class RequestMiddleware:
                     'method': request.method
                 },
             }, status=405)
+
+
+# Without this class the 'request revision' still works fine.
+# TODO: how to make it effective?
+# class BypassRevisionMiddleware(RevisionMiddleware):
+#
+#     def request_creates_revision(self, request):
+#         # Bypass the revision according to ...
+#         silent = request.META.get("HTTP_X_NOREVISION", "false")
+#         return super().request_creates_revision(request) and \
+#             silent != "true"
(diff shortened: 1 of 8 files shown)


middlewares/middlewares.py
```diff
@@ -67,3 +68,12 @@ def process_view(self, request, view_function, *args, **kwargs):
                     'method': request.method
                 },
             }, status=405)
+
+
+class BypassRevisionMiddleware(RevisionMiddleware):
```
szabozoltan69, line 73: Is it possible that this part is fully unnecessary?

thenav56:
@szabozoltan69 When we add a model as a revision in the admin panel, they are registered as revision models. So only those models will have a snapshot created by the revision.
https://django-reversion.readthedocs.io/en/stable/admin.html#registering-models

>  (Strange that there is "true", "false" in the code, instead of True, False.)
These attributes are passed from Headers `HTTP_X_NOREVISION` which only supports 'true'/'false'.

middlewares/middlewares.py
```diff
@@ -67,3 +68,14 @@ def process_view(self, request, view_function, *args, **kwargs):
...
+
+# Without this class the 'request revision' still works fine.
+# TODO: how to make it effective?
+# class BypassRevisionMiddleware(RevisionMiddleware):
+#
+#     def request_creates_revision(self, request):
+#         # Bypass the revision according to ...
+#         silent = request.META.get("HTTP_X_NOREVISION", "false")
+#         return super().request_creates_revision(request) and \
+#             silent != "true"
```
thenav56, line 81: This is to fulfill the custom requirements. For eg: this middleware allows the client to skip revision by passing a header. We don't need this for now, but if we have any custom needs, then this is needed.

szabozoltan69:
Solves https://github.com/IFRCGo/go-api/issues/1386

szabozoltan69:
Do you think @thenav56 it will be ok?

Issue #1402, opened 2022-04-19
[PROD] event-id is not a number


szabozoltan69:
I found that error: https://pastebin.com/zvQZJBMC
It means that deployments_qset.filter(deployment__event_deployed_to=event_id) did not receive an integer number, but 4660additional-information.

This /api/v2/deployment/aggregated?event=4660additional-information request come from Googlebot, and maybe not really important, but maybe we could make an effort to filter out, when event_id is not an integer. In this case we could send an error message or put a # after the digits, or cut the string after the digits.

The impact is small, so it is not a severe issue, but would be nice to avoid these parts in error log, if possible.

szabozoltan69:
(Sometimes it occurs with "encoded hash mark", like:
```
ValueError: invalid literal for int() with base 10: '3972#details'

==> /home/ifrc/logs/access.log <==
 - - [17/Apr/2022:16:50:18 +0000] "GET /api/v2/deployment/aggregated?event=3972%23details HTTP/1.0" 500 27 "https://go.ifrc.org/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36"
```
)

thenav56:
@szabozoltan69 We can use django-filters with drf view which can validate the params. Or just use the django-filter to parse the parameters and filter the queryset.
Eg: https://github.com/the-deep/server/blob/6f47333e0b09479fb83d9bd6d70c01e721eec125/apps/entry/filter_set.py#L483

PR #1412, opened 2022-04-22
Fix via coreapi/from-documenting-your-api

szabozoltan69:
https://www.django-rest-framework.org/coreapi/from-documenting-your-api/#installation : adding public=False


szabozoltan69:
Solves https://github.com/IFRCGo/go-api/issues/1402

PR #1423, opened 2022-04-27
Stronger password policy: Aa1@ is mandatory.

szabozoltan69:
And minimum 8 chars passwords are required.


szabozoltan69:
Solves https://github.com/IFRCGo/go-api/issues/1424

PR #1429, opened 2022-05-02
Adding other fields

szabozoltan69:
Solves https://github.com/IFRCGo/go-frontend/issues/2098


api/models.py
```diff
@@ -1229,7 +1238,7 @@ class FieldReport(models.Model):
     districts = models.ManyToManyField(District, verbose_name=_('districts'), blank=True)
     countries = models.ManyToManyField(Country, verbose_name=_('countries'))
     regions = models.ManyToManyField(Region, verbose_name=_('regions'), blank=True)
-    status = models.IntegerField(verbose_name=_('status'), default=0)
+    status = models.IntegerField(choices=FieldReportStatus.choices, verbose_name=_('type'), default=0)
```
thenav56, line 1241:
```suggestion
    status = models.IntegerField(choices=FieldReportStatus.choices, verbose_name=_('status'), default=0)
```

api/models.py
```diff
@@ -1207,6 +1208,14 @@ def __str__(self):
 class FieldReport(models.Model):
     """ A field report for a disaster and country, containing documents """
 
+    class FieldReportStatus(IntegerChoices):
```
thenav56, line 1211:
```suggestion
    class Status(IntegerChoices):
```

This should be fine. Since this is under FieldReport. Also looks good while using :D FieldReport.Status.EW

api/models.py
thenav56: Oh.. okay. Maybe add a NOTE for that.

api/models.py
szabozoltan69: Done.

thenav56 (approved):

PR #1433, opened 2022-05-23
Feature/django 2.2 to 3.2

szabozoltan69:
Django 2.2 to Django 3.2 bump up.
To be tested carefully.
(Unit tests are OK, but in some cases fixed manually.)


szabozoltan69:
@thenav56 @batpad Do you have any ideas what shall I do with changed migration files?
Some of the old ones needed enumfields, so I had to remove them – but if I change old migration files, Django will think that they have been run already. How do we solve such cutover situations?

thenav56:
> @thenav56 @batpad Do you have any ideas what shall I do with changed migration files? Some of the old ones needed enumfields, so I had to remove them – but if I change old migration files, Django will think that they have been run already. How do we solve such cutover situations?

@szabozoltan69 It should be okay to change the enum fields to integer fields as they are the same on the DB level.
Similar to this https://github.com/IFRCGo/go-api/pull/1491/files

Issue #1452, opened 2022-06-15
[PROD] Subscription/Record types inconsistency


szabozoltan69:
The index_and_notify job follows the creation of entities, and indexes them + sends notifications for subscribers.

We use multiple lists for these entities which are not fitting to each other perfectly.
The problem comes from the fact that in the past we refactored these lists and we tried to introduce new types instead of the old ones: https://github.com/IFRCGo/go-api/commit/6acbab66ae231d822a603f66c60cd4eaadf725e4

The index_and_notify job uses this:
https://github.com/IFRCGo/go-api/blob/develop/api/management/commands/index_and_notify.py#L41
On frontend side we consider these subscriptions types:
https://github.com/IFRCGo/go-frontend/blob/develop/src/root/views/account.js#L74
On backend side we use these subscription types:
https://github.com/IFRCGo/go-api/blob/develop/notifications/models.py#L104
https://github.com/IFRCGo/go-api/blob/develop/notifications/models.py#L175

We should find a consolidation for all these.

In reality we have only these kind of subscriptions in the prod database:
```
rtype	count   name
 3	 589  surgeAlert
 4	1337  country
 5	1529  region
 6	5494  disasterType
 7	  35  PER due date
 8	1382  followedEvent
 9	 336  surge DEPLOYMENT_MESSAGES
10	 191  surge APPROACHING_END_OF_MISSION
11 	 334  weeklyDigest
12	 461  newEmergencies
13	 416  newOperations
14	 253  GENERAL_ANNOUNCEMENTS

select count(distinct user_id) from notifications_subscription;
```
The number of the users (on prod) who have any kind of subscriptions: 1425

PR #1453, opened 2022-06-15
Notifications models rtype-map sync

szabozoltan69:


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1452

PR #1484, opened 2022-07-27
Optimizing EmergencyProjectViewSet

szabozoltan69:
The /api/v2/emergency-project/ endpoint is quite resource-intensive. Keeps minutes until it gets all the data.

Some experiments to make it faster:
- to put into EmergencyProjectViewSet's prefetch_related part all the entities that is called (it means 'activities', 'deployed_eru', 'modified_by' in plus)
- to avoid fields = '\_\_all\_\_', but using the needed ones.
Fortunately MiniEventSerializer is used here, which does not contain FR references (as the Surge optimization bottleneck did).


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1485

PR #1486, opened 2022-08-03
Indexing also NS for sharpening visibility

szabozoltan69:


szabozoltan69:
Seems to be working, at least on frontend side (during search, with some console.log) we can see the "ns" entities as a space separated list:
kép

PR #1489, opened 2022-08-05
Fix for PDF export

szabozoltan69:


szabozoltan69:
Deployed to Production.

PR #1515, opened 2022-09-15
Fix issue with the budget file reading.

thenav56:


frozenhelium:
@szabozoltan69 this one too. hopefully this is the last one

PR #1517, opened 2022-09-16
Fix azure file read/open issue.

thenav56:
NOTE:
We should think about moving to a more stable/tested azure storage library for Django.
https://django-storages.readthedocs.io/en/latest/backends/azure.html

### DIFF
--- a/api/storage.py
+++ b/api/storage.py
@@ -73,7 +73,7 @@ class AzureStorage(Storage):
             blob_name=name
         )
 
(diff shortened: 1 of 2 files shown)


frozenhelium:
@szabozoltan69 please do the honors!

szabozoltan69:
Hopefully the Django 3.2 will be such.

szabozoltan69:
Deployed to Staging

PR #1521, opened 2022-09-16
Script to import admin2

Rub21:
@geohacker @batpad


geohacker:
Thank you for your work on this @Rub21! This looks good. We'll make this part of the workflow once we decide an initial data source.

Issue #1547, opened 2022-10-05
PROD: Flash update notification emails are not sent


tovari:
Sharing Flash updates to Donors and Donor groups are not sent on Prod:
![image](https://user-images.githubusercontent.com/68283148/194060364-84278d7f-4374-49ed-837a-15171f91235f.png)

Use the Share button and select Internal Admin donor group to share:
https://go.ifrc.org/flash-update/6/

It works on STAG:
![image](https://user-images.githubusercontent.com/68283148/194060174-1dd43fe9-b1c4-4e03-8efa-8099d80d59d9.png)

https://go-stage.ifrc.org/flash-update/26/

szabozoltan69:
On PROD I can see this in the docker log of celery-default:

```
[2022-10-05 11:41:43,599: INFO/MainProcess] Task flash_update.tasks.share_flash_update[4afedc34-c061-43fb-a3c3-eb533ead4b83] received
[2022-10-05 11:41:43,848: ERROR/ForkPoolWorker-1] Task flash_update.tasks.share_flash_update[4afedc34-c061-43fb-a3c3-eb533ead4b83] raised unexpected: TypeError('sequence item 0: expected str instance, set found')
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 450, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/celery/app/trace.py", line 731, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/ifrc/go-api/flash_update/tasks.py", line 37, in share_flash_update
    send_notification(
  File "/home/ifrc/go-api/notifications/notification.py", line 109, in send_notification
    recipients_as_string = ','.join(to_addresses)
TypeError: sequence item 0: expected str instance, set found
```

To compare with the Staging site log:
```
[...]

PR #1549, opened 2022-10-06
Flash update multiple addressees 2

szabozoltan69:


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1547

PR #1555, opened 2022-10-18
Add optimistic lock in dref

k9845:
Addresses xxxxxx 

## Changes

* ...

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


dref/serializers.py
```diff
@@ -348,6 +348,7 @@ def create(self, validated_data):
     def update(self, instance, validated_data):
         validated_data['modified_by'] = self.context['request'].user
         is_assessment_report = validated_data.get('is_assessment_report')
+        modified_at = validated_data.get('modified_at')
```
frozenhelium, line 351: Use pop, use None as default value

dref/serializers.py
```diff
@@ -373,18 +374,21 @@ def update(self, instance, validated_data):
...
-                lambda: send_dref_email.delay(dref.id, list(to), 'Updated')
-            )
-        return dref
+        if modified_at and instance.modified_at and modified_at > instance.modified_at:
+            dref = super().update(instance, validated_data)
+            if to:
+                transaction.on_commit(
+                    lambda: send_dref_email.delay(dref.id, list(to), 'Updated')
+                )
+            return dref
```
frozenhelium, line 390: Throw error if
- modified_at is None, 'modified_at is required'
- modified_at != instance.modified_at, 'OBSOLETE_PAYLOAD'
Note: defined 'OBSOLETE_PAYLOAD' as a static value, client will use this string to identify this type of error. Please add a note in code for this as well.

dref/test_views.py
```diff
@@ -847,3 +848,27 @@ def test_dref_for_super_user(self):
...
+        )
+        url = f'/api/v2/dref/{dref.id}/'
+        data = {
+            'title': "New title",
+            'modified_at': datetime(2022, 2, 18, 2, 29, 39, 793615),
+        }
+
+        self.client.force_authenticate(self.user)
+        response = self.client.patch(url, data=data)
+        self.assertEqual(response.status_code, 200)
```
frozenhelium, line 866: This should be 400, also assert for the error response
Add a test case to verify `modified_at` is not null

frozenhelium (changes requested):

dref/models.py
```diff
@@ -227,7 +228,10 @@ class Status(models.IntegerChoices):
         COMPLETED = 1, _('Completed')
 
     created_at = models.DateTimeField(verbose_name=_('created at'), auto_now_add=True)
-    modified_at = models.DateTimeField(verbose_name=_('modified at'), auto_now=True)
+    modified_at = models.DateTimeField(
+        verbose_name=_('modified at'),
+        default=datetime.now, blank=True
```
thenav56, line 233: Use auto_now instead of datetime.now
https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.DateField.auto_now

PR #1557, opened 2022-10-19
admin2 imports

geohacker:
This PR prepares the admin2 import script.


api/management/commands/import-admin2-data.py
```diff
@@ -77,13 +71,18 @@ def handle(self, *args, **options):
 
       # loop through each feature in the shapefile
       for feature in data[0]:
-          code = feature.get("code")
-          name = feature.get("name")
+          code = feature.get("code") if 'code' in feature.fields else feature.get("pcode")
```
batpad, line 74: @geohacker can there be a case where there's a `code` and a `pcode` and this would cause `pcode` to get dropped? Is that an issue? Should we store `pcode` as a separate field to make sure that is always a `pcode`?

api/snapshots/snap_test_views.py
```diff
@@ -11,63 +11,63 @@
     "dtype": 1,
     "countries": [],
     "districts": [],
-    "summary": "MGyDLJYVcCZKPmuMEGjdCgZvTfGPlcpTCCHHNkxxsyAXvRMdYOPvevgJRysqUQMjvfLQjwtPSQziMTftJyPYviQSVRHfPQBGxbxtlnvXFmoijesYgGXIVHcQvXNiMyjklSXNZkUCcAxRUpCNsWVYCoIptZYEmxRKCDXsXyGHAkmZMiqdPExJgTHhsfWkrCGjBfoCwbAdzGxpyfxobugTPvYjicsESiWTECNafbqnjJUMHBhXspthdpAOYNDehFMIbOGKpTjsBaNwpKAlQQfHxeHIGYGJbyEcOyxqVbwYewpUQOgXLVWvicwIvPlXRDSEOlZieTXDcsmcYmcutGzIEqcWPmswXdPvrhZxBzVCyvlFSFxZHrZfUBfBMlIsugfuQstCMTBkSCwCcUwNBrOYdeQOzxGZVRkbjMRYCciepXPxxyKcMjRCxxCWeKiHxzuPrphbVlFHyJhqXqTCnNsSFmhieClTCfZRuQwTeJIstkTTSOlYxGoh",
+    "summary": "NMGyDLJYVcCZKPmuMEGjdCgZvTfGPlcpTCCHHNkxxsyAXvRMdYOPvevgJRysqUQMjvfLQjwtPSQziMTftJyPYviQSVRHfPQBGxbxtlnvXFmoijesYgGXIVHcQvXNiMyjklSXNZkUCcAxRUpCNsWVYCoIptZYEmxRKCDXsXyGHAkmZMiqdPExJgTHhsfWkrCGjBfoCwbAdzGxpyfxobugTPvYjicsESiWTECNafbqnjJUMHBhXspthdpAOYNDehFMIbOGKpTjsBaNwpKAlQQfHxeHIGYGJbyEcOyxqVbwYewpUQOgXLVWvicwIvPlXRDSEOlZieTXDcsmcYmcutGzIEqcWPmswXdPvrhZxBzVCyvlFSFxZHrZfUBfBMlIsugfuQstCMTBkSCwCcUwNBrOYdeQOzxGZVRkbjMRYCciepXPxxyKcMjRCxxCWeKiHxzuPrphbVlFHyJhqXqTCnNsSFmhieClTCfZRuQwTeJIstkTTSOlYxGo",
```
batpad, line 14: Not sure what's going on with these changes to the snapshot tests - none of these look like they should be changes caused by this PR? Can we remove these changes from the PR, and then look if tests fail on CI?

api/management/commands/import-admin2-data.py
geohacker: Ah good point. It is possible but our instructions are clear that it should be either code or pcode.

api/snapshots/snap_test_views.py
geohacker: @szabozoltan69 was kind to fix these tests for me in this PR so I think it's ok!

szabozoltan69: Hey, these snapshot tests are behaving like a "living plant". Time to time they change, based on date. Fortunately the needed new string is shown by the test result.
(If I were be able to eliminate these changes, I would do it.)

szabozoltan69:
There were some migration conflict, but after solving them I will deploy this to staging (in 10 minutes).

szabozoltan69:
@geohacker @batpad – Deployed to Staging.

PR #1558, opened 2022-10-19
Choices showing viewset

szabozoltan69:
@batpad @thenav56 If you have a better idea to show ERUTypes on an endpoint, tell me, please.
(This one still not works)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ email-verification only, is to be found
 ### Running server
 
      $ docker-compose up serve celery
-     $ (Obsolete: docker-compose run --rm --service-ports serve)
+     $ (Without celery, the old pdb-friendly way: docker-compose run --rm --service-ports serve)
 
 Access the site at http://localhost:8000
 
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ email-verification only, is to be found
 ### Running server
 
      $ docker-compose up serve celery
-     $ (Without celery, the old pdb-friendly way: docker-compose run --rm --service-ports serve)
+     $ (To attach to a container with stdin, e.g. for debugging: docker attach <container_name>)
+     $ (Or, without celery, the old pdb-friendly way: docker-compose run --rm --service-ports serve)
 
 Access the site at http://localhost:8000
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ email-verification only, is to be found
 
 ### Running server
 
-     $ docker-compose run --rm --service-ports serve
+     $ docker-compose up serve celery
+     $ (Obsolete: docker-compose run --rm --service-ports serve)
 
 Access the site at http://localhost:8000
 
--- a/deployments/drf_views.py
+++ b/deployments/drf_views.py
@@ -30,6 +30,7 @@ from api.visibility_class import ReadOnlyVisibilityViewsetMixin
 from .filters import ProjectFilter, EmergencyProjectFilter
 from .utils import get_previous_months
 from .models import (
+    ERUType,
     ERU,
     ERUOwner,
     OperationTypes,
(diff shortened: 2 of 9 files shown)


deployments/drf_views.py
```diff
@@ -68,6 +70,17 @@
 )
 
 
+class ERUTypeViewset(viewsets.ReadOnlyModelViewSet):
```
batpad, line 73: Hmm this should likely not be a `ReadOnlyModelViewset` as this is not really returning results for a "model".

We just want a pretty simple view where we can get all the various enum options as JSON and return it.

@thenav56 so you think we can just make an entry in `urls.py` and point to a simple function based view for this? Or is there a class based view that seems appropriate?

(Essentially, we want this endpoint to just return a JSON giving all the values of all the ENUMs in the backend).

main/urls.py
```diff
@@ -154,6 +154,7 @@
 router.register(r'dref-op-update', dref_views.DrefOperationalUpdateViewSet, basename='dref_operational_update')
 router.register(r'dref-final-report', dref_views.DrefFinalReportViewSet, basename='dref_final_report')
 router.register(r'review-country', api_views.CountryOfFieldReportToReviewViewset, basename='review_country')
+router.register(r'erutype', deployment_views.ERUTypeViewset, basename='erutype')
```
batpad, line 157: We might not want to use the `router.register` here but instead just map a url to a view like https://github.com/IFRCGo/go-api/blob/develop/main/urls.py#L166 .. let's coordinate on teams and can sketch out a basic structure. cc @thenav56

README.md
```diff
@@ -48,7 +48,8 @@ email-verification only, is to be found
 
 ### Running server
 
-     $ docker-compose run --rm --service-ports serve
+     $ docker-compose up serve celery
+     $ (Without celery, the old pdb-friendly way: docker-compose run --rm --service-ports serve)
```
thenav56, line 52: @szabozoltan69 Maybe we can use this as well.

https://github.com/the-deep/deeper/blob/35f01f977293d1c26d0667bd1ea6458dcf4e4397/docker-compose.yml#L7-L10

deployments/drf_views.py
thenav56: Yeah, a simple view should suffice for this.

Eg.
url: /enum-options/
payload:
```json
{
    "NationalSocietyActionTitle": [
        {
            "key": "national_society_readiness",
            "label": "National Society Readiness"
        },
        ...
    ],
    ...
}
```

utils/erp.py
```diff
@@ -60,6 +60,15 @@ def push_fr_data(data, retired=False):
...
         InitialRequestType, InitialRequestValue =  "Empty", 0
 
+    # In Early Warning / Early Action we use "Potentially Affected", not the normal _affected ones:
+    NumberOfPeopleAffected = \
+        (0 if data.num_affected is None else data.num_affected) + \
+        (0 if data.gov_num_affected is None else data.gov_num_affected) + \
+        (0 if data.other_num_affected is None else data.other_num_affected) + \
+        (0 if data.num_potentially_affected is None else data.num_potentially_affected) + \
+        (0 if data.gov_num_potentially_affected is None else data.gov_num_potentially_affected) + \
+        (0 if data.other_num_potentially_affected is None else data.other_num_potentially_affected)
```
thenav56, line 70: We can also do this.
```suggestion
    number_of_people_affected = sum([
        affected for affected in [
            data.num_affected,
            data.gov_num_affected,
            data.other_num_affected,
            data.num_potentially_affected,
            data.gov_num_potentially_affected,
            data.other_num_potentially_affect,
        ]
        if affected is not None
    ])
```

szabozoltan69:
@batpad @thenav56 thanks for the lots of help! It was eye-opener for me.

I merged and deployed the recent one to staging. We can sophisticate this in the future and do it for other "choice" types:
https://dsgocdnapi.azureedge.net/api/v2/erutype

Issue #1532, opened 2022-09-20
Standardize validation errors returned by the backend


batpad:
For forms like the 3W form, we use DRF classes to handle serializing and saving data. This lets us leverage DRF's standardized error reporting, and DRF then returns validation errors in a standardized format: https://www.django-rest-framework.org/api-guide/exceptions/#validationerror

For some forms, we don't use DRF views / serializers to save the data, and there error reporting does not follow the DRF conventions. This adds a lot of work to the frontend to deal with multiple non-standard ways of representing validation errors from the backend.

We should standardize / ensure all backend validation error responses for form saving follows the DRF format for representing errors and correctly return a `400` status code.

TODO:

  - @frozenhelium to help enumerate all forms where the backend does not return errors in the standard format
  - Then, @thenav56 and @szabozoltan69 can work together to ensure they all return errors in the DRF format, ideally by converting those views to use DRF serializers.

cc @tovari @tnagorra

szabozoltan69:
@batpad @thenav56 Should we rewrite this? https://github.com/IFRCGo/go-api/blob/develop/api/exceptions.py#L4-L8

szabozoltan69:
Also https://github.com/IFRCGo/go-api/pull/1715 lessons learned should be canalized to this solution.
So: login, FR, account profile data changes should handle backend errors like this:
```
'errors': {
     'non_field_errors': ['Hey, we are in maintenance mode, come back a bit later – site is in read only mode']
}
```

szabozoltan69:
Solved via web-rewrite

PR #1570, opened 2022-10-27
Replacement of bad_request with field-focused bad_form_request

szabozoltan69:


api/views.py
```diff
@@ -40,6 +40,12 @@ def bad_http_request(header, message):
     return HttpResponse('<h2>%s</h2><p>%s</p>' % (header, message), status=400)
 
 
+def bad_form_request(field_name, message, statusCode=400):
+    return JsonResponse({'errors': [{
+        'field_name': field_name,
+        'error_message': message}]}, status=statusCode)
+
```
thenav56, line 47: @frozenhelium Is this format okay for the client-side?

thenav56 (approved):
LGTM

api/views.py
batpad: I don't think this is the exact format we want. Let's hold on merging until we can discuss this with @frozenhelium and @tnagorra - but, yes, hopefully that's just a matter of changing the JSON structure in this one function - otherwise this looks good!

tnagorra: This looks good for now.

batpad (approved):

szabozoltan69:
Thanks @batpad @tnagorra @thenav56 @frozenhelium !
(I still miss that feature that at once multiple field can be "shown" as erroneous.)
Before merging, could you suggest me other .py files where the old style error messaging is still present?

szabozoltan69:
We should change frontend side also, because the old code gives this (in case of wrong password):
image

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1532

batpad:
> We should change frontend side also, because the old code gives this (in case of wrong password):

Yes, of course 🤦 . Let's hold off on merging this until @frozenhelium or @tnagorra can say what it would take to get the frontend to work with this style? I guess ideally this would mean moving these forms over to the new forms? But not sure how much work that could be. We should probably hold off on merging the backend change before we have the frontend work lined up.

szabozoltan69:
Frontend changes: https://github.com/IFRCGo/go-frontend/pull/2540

frozenhelium:
@szabozoltan69 I think this also works for now, but in my opinion we should generalize the field names a bit since the error message might not always be a string. My suggestion for now is:
```js
error = [
  {
    field_name: <field_name>,

    // Could be string or object or array depending on type of field
    error: <error>,
  },
];
```

szabozoltan69:
Thanks @frozenhelium – it's not quite clear how to create a generalized type in Python. (Or just simply we use the one that is appropriate at the different locations.)

tnagorra:
The client also needs to know what the type of the field is. So, either we have to have an additional field to indicate if the field is `Array`, `Object` or `Literal` or we have three different fields for their errors like `arrayError`, `objectError` and `literalError`

PR #1572, opened 2022-10-28
Revoke full permission from ifrc_admin

szabozoltan69:


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1573

PR #1575, opened 2022-11-02
admin2 vector tiles + better tile management commands

geohacker:
This PR:
* makes the tile update commands more reliable
* add support to update countries and districts separately if needed
* add support for creating and updating admin2 tiles
* add support for staging tiles (this is only available for admin2s right now because creating them for countries and districts would incur a cost)

cc @batpad @szabozoltan69


szabozoltan69 (approved):
I think it's great.
This is the most resource-consuming job in all GO platform, so we should test it on a "strong vm". I hope locally it worked fine in your environment, @geohacker – thanks for leaving the 1min waits.

batpad (approved):

PR #1577, opened 2022-11-03
Field reports figures into separate table

szabozoltan69:


tnagorra:
@szabozoltan69 Is this branch still needed?

szabozoltan69:
@tnagorra Well, yes, this task is still not done – and in case of "plenty of time" (or backend refactoring) we should solve it once. The lots of "num..." fields in FieldReport should not look like the recent solution.

tnagorra:
@szabozoltan69 should we convert this to draft PR then?

PR #1584, opened 2022-11-14
Terraform + Helm Deploy

geohacker:
WIP

cc @batpad


szabozoltan69:
@batpad @geohacker – I would have a few comments to this nice work.
One was already mentioned before, that let's use GO with uppercase, if possible. (IFRC Go Role Assigner - > IFRC GO...)
There are some variables which are used but not in the above list (and not in main/settings.py either), these are:
```
APPEALS_USER
APPEALS_PASS
```
https://github.com/IFRCGo/go-api/blob/develop/api/management/commands/ingest_appeals.py#L90
Maybe DB_SERVER_NAME would not be bad also, because it is needed in 2 other variables, so we could avoid redundance:
```
DJANGO_DB_HOST="$DB_SERVER_NAME.postgres.database.azure.com"
DJANGO_DB_USER="...@$DB_SERVER_NAME" (! needed here !)
```
If you would like to have a look at the "old-fashioned" scripts and variables, I can send them. Maybe a look-over would be not bad before finalizing the "new fashion".
(I wonder whether the lot's of redundancy of the same variable mentioning is not avoidable somehow... This way a new variable adding will be a nightmare. We should document it anyway, what will be the process of it.)

Issue #1586, opened 2022-11-15
[PROD] Only active users for DREF


szabozoltan69:
## Issue
The DREF page needs some users to be shared with. The API shows also inactive users. This should not be so. We should fix the related endpoint to list only active users. The inactive users should be treated as nonexistent ones.

szabozoltan69:
On dref-options endpoint ()
dref/views.py::DrefOptionsView
```
-                } for user in User.objects.all()
+                } for user in User.objects.filter(is_active=True).all()
```
addition will solve the issue.

PR #1587, opened 2022-11-15
Only active users are usable in DREF

szabozoltan69:


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1586

PR #1598, opened 2022-11-25
Fix country-plan MembershipCoordination issue.

thenav56:
Addresses Missing membership coordination for country plan.


szabozoltan69:
Deployed :-)

PR #1609, opened 2022-12-08
Prohibit language prepending in case of 404 not found

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1594

Issue #1594, opened 2022-11-25
[PROD] 404 in endpoint redirect to login page.


thenav56:
## Issue
*Currently there is a issue where all 404 errors are redirected to **/en/{url}** for a given **url**. This is also redirect next to the admin login page.*

## Steps to reproduce
*Goto into region endpoint with non-existing id.
`curl -isL https://goadmin.ifrc.org/api/v2/region/123123/`*

![2022-11-25-103448](https://user-images.githubusercontent.com/7059255/203903479-b8bbd9fe-885a-469d-86ee-2980ab2c9ac0.png)

## Expected behavior
*Respond with DRF standard 404 response.*

## Severity
*This affects the error handling on the front end side.*

## NOTE
Maybe this is related to https://code.djangoproject.com/ticket/17734

szabozoltan69:
Nice catch!
Could be reproduced locally also.
The source of this redirection (to /en/... ) is not yet found.

szabozoltan69:
If I set this variable from True to False, this bug disappears. But then it is not possible to change the language of the admin page:
https://github.com/IFRCGo/go-api/blob/develop/main/urls.py#L245

szabozoltan69:
This way the bug is solved. So we should somehow subclass this !!! part in 
/usr/local/lib/python3.8/site-packages/django/middleware/locale.py :

 ```
from django.utils.deprecation import MiddlewareMixin


class LocaleMiddleware(MiddlewareMixin):
    """
    Parse a request and decide what translation object to install in the
    current thread context. This allows pages to be dynamically translated to
    the language the user desires (if the language is available).
    """
    response_redirect_class = HttpResponseRedirect

    def process_request(self, request):
        urlconf = getattr(request, 'urlconf', settings.ROOT_URLCONF)
        i18n_patterns_used, prefixed_default_language = is_language_prefix_patterns_used(urlconf)
        language = translation.get_language_from_request(request, check_path=i18n_patterns_used)
        language_from_path = translation.get_language_from_path(request.path_info)
        if not language_from_path and i18n_patterns_used and not prefixed_default_language:
            language = settings.LANGUAGE_CODE
        translation.activate(language)
        request.LANGUAGE_CODE = translation.get_language()

    def process_response(self, request, response):
        language = translation.get_language()
        language_from_path = translation.get_language_from_path(request.path_info)
        urlconf = getattr(request, 'urlconf', settings.ROOT_URLCONF)
[...]

PR #1612, opened 2022-12-12
alternate approach to admin redirect

batpad:
@szabozoltan69 this was my idea to handle the `admin` redirect - can you let me know if this works? I'm slightly confused with the handling of the `language` parameter - but I think this should be a cleaner separation of the admin URLs than what we have right now.

### DIFF
--- a/main/urls.py
+++ b/main/urls.py
@@ -219,12 +219,12 @@ urlpatterns = [
     url(r'^api/v2/exportperresults/', per_views.ExportAssessmentToCSVViewset.as_view()),
     url(r'^docs/', include_docs_urls(title='IFRC GO API', public=False)),
     url(r'^tinymce/', include('tinymce.urls')),
-    url(r'^admin/', RedirectView.as_view(url='/')),
     # url(r'^', admin.site.urls),
     url(r'^favicon\.ico$', RedirectView.as_view(url='/static/favicon.ico')),
     url(r'^server-error-for-devs', DummyHttpStatusError.as_view()),
     url(r'^exception-error-for-devs', DummyExceptionError.as_view()),
     path('i18n/', include('django.conf.urls.i18n')),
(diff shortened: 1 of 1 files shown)


batpad:
cc @thenav56

szabozoltan69:
This solution is great. Works perfectly for me.
Also the original https://github.com/IFRCGo/go-api/issues/1594 issue is solved:
```
curl -isL 'http://localhost:8000/api/v2/region/9/'
HTTP/1.1 404 Not Found
Date: Mon, 12 Dec 2022 15:45:22 GMT
Server: WSGIServer/0.2 CPython/3.8.12
Content-Type: application/json
Vary: Accept, Accept-Language, Origin, Cookie
Allow: GET, HEAD, OPTIONS
X-Frame-Options: DENY
Content-Length: 23
Content-Language: en
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin

{"detail":"Not found."}
```
This is a great solution because this way the [process_response of site-packages/django/middleware/locale.py](https://github.com/django/django/blob/main/django/middleware/locale.py#L72)
works fine, path_valid remains false in case of erroneuos ID request.

szabozoltan69:
Before merging we should consider that there are some points on frontend where we turn the user to the admin page, e.g. Emergency edit. These should be also aligned with this change (inserting /admin into the URL).

thenav56:
Totally agree with these changes. This will also solve the issue with the language redirect fixed by https://github.com/IFRCGo/go-api/pull/1609.

I am just concerned about the usage of the admin URLs if it is shared with others.

batpad:
@szabozoltan69 sorry for the delay here! Maybe let's not push into this production release, but immediately after we can merge this to develop, make sure we test well on staging and then get this into the next release?

szabozoltan69:
Good approach!
And it has some frontend task also ("Edit Event" link to be changed).

Issue #1535, opened 2022-09-22
Add branch offices data to the API - Local unit mapping


geohacker:
This will involve:
* [ ] Inspecting the data @jhenshall @geohacker 
* [ ] Create migrations for new models
* [ ] Management command for importing data from shp/csv
* [ ] API routes
* [ ] Test

@batpad @szabozoltan69 @thenav56  @tovari @LukeCaley

geohacker:
<img width="929" alt="image" src="https://user-images.githubusercontent.com/371666/191658949-3a25fd4c-388e-4aa8-b37e-175da0b1d4bf.png">

The data _looks_ ok but there are problems. Overall:
1. There are 2584 points
2. It's unclear what NS_Type, NS_SubType. These have several options that looks not very clean `1,2,3,4, some lat/lng values, IFRC` etc

What attributes do we want to import?
1. Country ISO (2 or 3)
2. National Society name
3. Branch name 
4. Branch type (we don't have this now. NS/IFRC/Sub-branch?) 
5. Created timestamp
6. Last updated timestamp
7. Draft boolean (default false)
8. Validated boolean (default true)
9. Source? (NS, ICRC, Others?)
10. Sensitive / Public ?
12. Address

esario:
I have an up-to-date highly accurate dataset of Finnish RC branch data. The issue is that there's no point data, only polygons (branch polygons are comprised of 1-n zip code areas). I can compute centroids, but for attributes, there is no address and postcode data - appreciate any advice on how to tackle.

geohacker:
@esario do you mind sharing that data with me? We can handle computing centroids during the import. For address and postcode, I'd suggest we leave them empty for now and then when you have it either go through the GO Admin to edit or reimport if that's too slow a process.

geohacker:
To move forward on this, we decided in Kathmandu that we'd start with Ukraine data that looks reasonably complete.

batpad:
cc @willemarcel here - it would be great if you could start looking at this. I think what we'd need roughly:

 - [ ] A new model for `NSBranchOffice` or so that would have a ForeignKey to `Country` and the attributes listed above (@geohacker please confirm)
 - [ ] Add API endpoints to LIST Branch offices, with some filters (perhaps just `country` for now), and GET details of a branch office.
 - [ ] Ensure that Admin interface to edit Branch offices work, including ability to move point on the map.
[...]

PR #1622, opened 2022-12-22
Add LocalUnit Model & endpoints

willemarcel:
Addresses #1535 

## Changes

* Add LocalUnit model, list and detail endpoints, admin interface and tests


willemarcel:
@batpad it's missing only the import scripts. Let me know if it needs some change, specially in the model.

batpad:
@willemarcel this looks really good to me!

@geohacker - could you please take a look at the model definition and let know if the fields look good?

local_units/models.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    modified_at = models.DateTimeField(
+        verbose_name=_('Updated at'),
+        auto_now=True
+    )
+    draft = models.BooleanField(default=False, verbose_name=_('Draft'))
+    validated = models.BooleanField(default=True, verbose_name=_('Validated'))
+    source = models.CharField(max_length=255, verbose_name=_('Source'))
+    address = models.CharField(max_length=500, verbose_name=_('Address'))
+    post_code = models.CharField(max_length=10, verbose_name=_('Postal code'))
```
geohacker, line 49: Let's make postcode a single word

geohacker:
@willemarcel thank you for this! This PR looks great to me. @davidmuchatiza shared a sample data for the local units (I'm forwarding this to you). With that I think we need to make some small adjustments here — mostly adding new columns. 

1. address_loc and address_en
2. city_loc and city_en
3. focal_person_loc and focal_person_en


@davidmuchatiza you have a `GO ID` column. Is that meant to be the country id from GO?

cc @batpad

davidmuchatiza:
> @willemarcel thank you for this! This PR looks great to me. @davidmuchatiza shared a sample data for the local units (I'm forwarding this to you). With that I think we need to make some small adjustments here — mostly adding new columns.
> 
> 1. address_loc and address_en
> 2. city_loc and city_en
> 3. focal_person_loc and focal_person_en
> 
> @davidmuchatiza you have a `GO ID` column. Is that meant to be the country id from GO?
> 
> cc @batpad

Yes sure thus GO Country ID, to allow us to link the dataset to the country table.

willemarcel:
@batpad @geohacker I have finished the changes in the model and added the import CSV command.

batpad:
@willemarcel I'm not seeing the Import CSV command in this PR - maybe you forgot to commit a file or so?

willemarcel:
@batpad Sorry, I forgot to push the commit yesterday.

local_units/management/commands/import-local-units-csv.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        filename = options['filename'][0]
+        with open(filename) as csvfile:
+            reader = csv.DictReader(csvfile)
+            for row in reader:
+                unit = LocalUnit()
+                unit.country = Country.objects.get(iso3=row['ISO3'])
+                unit.type = LocalUnitType.objects.get_or_create(
```
batpad, line 25: @willemarcel can we do a `print` here if a new type was created? I think it will be useful to warn / inform users when a new type is created so that if it was created accidentally / due to some sort of spelling mistake / typo, they are able to manually correct.

willemarcel: @batpad I've added it

batpad:
@willemarcel this looks great!

We should probably clarify the best workflow around handling types in the spreadsheet imports - it would be great to get any thoughts or suggestions you have for how we should ideally structure the CSV for types. But for now, this code looks good to me.

The test failures are weird :-/

Would really like to figure out what's causing that 302 on Circle causing one of your tests to fail. If it passes locally, not sure what to do about this one.

For the failing snapshot test, that looks unrelated to your changes - @szabozoltan69 - do you think you'd be able to help here?

szabozoltan69:
Yes, I'll fix it soon, these are "time-to-time to-be-rewritten" snapshot test errors.
(Otherwise, if you would like to fix it yourself, the "+" rows are in the code, they should be replaced with the "-" rows.
E.g. instead of country-FWJNPIGxPkqKRLeNzbOHsgUSwRJXvxmDTodTYrnDAnhHuxKzFb the new string is
country-hkFWJNPIGxPkqKRLeNzbOHsgUSwRJXvxmDTodTYrnDAnhHuxKz, and the test will work again. I don't know exactly the reason of the forced changes, but they may be due to date change.)

szabozoltan69:
I commented the "/api/v2/local-unit/1/" asserts to have a working test

szabozoltan69:
@willemarcel @batpad do you read it? :-)

batpad:
Yea, commenting out the test in views.py is non-ideal :( let's maybe give some effort to fixing that while keeping the test.

@szabozoltan69 if you keep those lines in views.py, does the test pass for you locally or is it failing for you locally?

szabozoltan69:
@batbad I did not try it, just I saw in https://app.circleci.com/pipelines/github/IFRCGo/go-api/3269/workflows/5fb64e9c-8b43-4608-a18d-023ff4633ec8/jobs/5028 that 
```
 def test_detail(self):
        response = self.client.get('/api/v2/local-unit/1/')
>       self.assertEqual(response.status_code, 200)
E       AssertionError: 302 != 200
```
... but as it is not related to snapshot test auto-changing, I left it as it is (but commented, not to have broken test).

szabozoltan69:
So I would like to ask the writer of that test to fix it. :-)

batpad:
@szabozoltan69 are you able to just try running the test locally though? The problem is @willemarcel is unable to reproduce the failing test locally (it passes locally but fails on Circle, somehow :( ) - I will also try this locally tomorrow. But would just be good to know if for others it works locally and the issue is only Circle or if anyone else can reproduce this locally.

szabozoltan69:
Trying to do it... (a bit time consuming).

szabozoltan69:
Locally it was running fine for me also.
```
=== 98 passed, 1107 warnings in 352.75s (0:05:52) ====
```

Otherwise it is not fully clear for me why do we assume that exactly the /1 will exist in the test database...
```
        LocalUnitFactory.create_batch(2, country=country, type=type)

    def test_detail(self):
        response = self.client.get('/api/v2/local-unit/1/')
```

szabozoltan69:
The working  '.../local-unit/X' should not be fixed.
With a pdb.set_trace() into https://github.com/IFRCGo/go-api/blob/feature/local-units-app/local_units/views.py#L25 I discovered that this X id increases always.
So, for me once response = self.client.get('/api/v2/local-unit/81/') worked fine:
```
![image](https://user-images.githubusercontent.com/6050643/212147976-20163af5-e380-44f6-ab4f-ecca88613bea.png)
But later the same id broke the test.

batpad:
WoW! @szabozoltan69 thank you SO MUCH for digging into this.

@willemarcel can we try not hard-coding the id in the test but rather make a query to the database to get the id of the first element or so?

So before this line: https://github.com/IFRCGo/go-api/pull/1622/files#diff-d001e4b5d38843f23c7966b1ec20886c072721360ed09fb0bf138fffd3ba2814R58

Maybe just like `id = LocalUnit.objects.all()[0].id` or so and then use that id instead of assuming the id will be `1` ?

Am still a bit confused about why Circle would do this, but this kindof makes sense and am hoping this can fix the issue.

Thanks so much again @szabozoltan69 !

szabozoltan69:
@batpad what do you think, can we put this feature into the today's release (even without the erroneous tests, which can be fixed later)?

batpad:
@szabozoltan69 no - let's wait - anyways, there is no "feature" for users with this code -- we would need to merge this to develop, then test some imports of CSVs on staging, etc. so no hurry to get this onto production..

willemarcel:
@batpad @szabozoltan69 Thanks for the help! It's ready now!

tovari:
Not strictly related to this, but can we use the geocoding service of mapbox, if we have only address info for the location @geohacker?

geohacker:
@tovari we could try but I'm not sure how accurate / reliable it will be for practical use.

PR #1652, opened 2023-02-07
Add issues to Backlog

szabozoltan69:
@batpad @tovari 
(If works, the similar should has been done with go-frontend and go-user-library.)


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1653

Issue #1647, opened 2023-02-03
Changes / fixes to Local Units implementation


batpad:
Some minor things that would be good to fix / improve with the implementation for Local Units:

 - Add a `__str__` method to ensure it shows up nicely in the Admin -- maybe can be something like `local_branch_name` + `country.name` or so?
 - Allow filtering in the API. One should be able to filter by `country` ISO code and `type`. We can consider other filters, including geospatial, later.

@szabozoltan69 anything else you can think of that can be improved with the local units import? Should we open another ticket about error handling in the CSV import script?

@willemarcel do you think you'd be able to take on the filtering in the API and the `__str__` method to have a nice display name in the admin list?

szabozoltan69:
Just one comment: the .csv file should not contain BOM character as start (otherwise the reader does not find the ISO3 column, as it is found as ¤ISO3).

szabozoltan69:
Deployed to Staging

PR #1673, opened 2023-02-09
Add __str__ method to LocalUnit model

willemarcel:
Contributes to #1647

## Changes

* Add __str__ method to LocalUnit model, returning the branch and country names


szabozoltan69:
Could you please, @willemarcel, create some similar naming for LocalUnit types?
The recent view is this:
kép

szabozoltan69:
I think this would be fine:
```
    def __str__(self):
        return f'{self.name} ({self.level})'
```

PR #1692, opened 2023-02-28
Adding /media to .dockerignore

szabozoltan69:


.dockerignore
```diff
@@ -1 +1,2 @@
 .db/
+media/
```
thenav56, line 2: Hi @szabozoltan69 

In my local, I am also seeing these files. Add them as well if this exists on yours.
```
.mypy_cache
.pytest_cache
go-logs
go.log
```

thenav56 (approved):

Issue #1699, opened 2023-03-06
[PROD] Project endpoint - too much queries


szabozoltan69:
## Issue
There are incredible quantity of unnecessary queries on backend - if user is logged in.
Was discovered via DjagoDebugToolbar while implementing https://github.com/IFRCGo/go-api/issues/1693 – but it was also actual before this change.

.../api/v2/project/
not logged in:2336.08 ms (3626 queries including 3617 similar and 3554 duplicates )
logged in:5479.40 ms (8169 queries including 8155 similar and 8142 duplicates )

.../api/v2/project/198/
not logged in:20.18 ms (7 queries )  !!! GOOD !!! The others should be like this also
logged in:2965.70 ms (4555 queries including 4541 similar and 4501 duplicates )

Mainly this kind of queries are repeated:
SELECT ••• FROM "api_country" WHERE "api_country"."id" = 14 LIMIT 21
4539 similar queries. Duplicated 68 times.

szabozoltan69:
If you have any idea, please tell it, dear @thenav56

batpad:
This is a really good catch and seems like it should be just missing `select_related`, `prefetch_related` type stuff somewhere? @thenav56 do you think you'd be able to look?

thenav56:
Hi there, I looked around and most of the query is due to the form rendered in the DRF `HTML form`.

Basically, due to this line, https://github.com/IFRCGo/go-api/blob/develop/api/models.py#L261 Django has to fetch country causing N+1 query.


There is one prefetch missing from the project. This diff should fix for now. @szabozoltan69 Could you add this to your existing PR?
```diff
diff --git a/deployments/drf_views.py b/deployments/drf_views.py
index 38f13638..3ba40ee9 100644
--- a/deployments/drf_views.py
+++ b/deployments/drf_views.py
@@ -373,7 +373,7 @@ class ProjectViewset(
     viewsets.ModelViewSet,
 ):
     queryset = Project.objects.prefetch_related(
-        'user', 'reporting_ns', 'project_districts', 'event', 'dtype', 'regional_project', 'annual_splits',
+        'user', 'project_country', 'reporting_ns', 'project_districts', 'event', 'dtype', 'regional_project', 'annual_splits',
     ).all()
     filterset_class = ProjectFilter
     serializer_class = ProjectSerializer
```

Tip:
[...]

PR #1701, opened 2023-03-07
Protect Sector at deletion. Other cleanups

szabozoltan69:


szabozoltan69:
This issue https://github.com/IFRCGo/go-api/issues/1699 is also to be solved via this change.

PR #1715, opened 2023-03-20
Django read only for maintenance mode

szabozoltan69:


szabozoltan69:
Maybe a better feedback could be done via frontend. In case of write attempt: something like "hey, we are in maintenance mode, come back a bit later – site is in read only mode".

thenav56:
@szabozoltan69 
Maybe we can update our error handler to process the DjangoReadOnlyError expectation and return a standard message using `non_field_errors` which then can be handled by frontend form components.

GO-API Error Handler: https://github.com/IFRCGo/go-api/blob/develop/main/exception_handler.py#L14

szabozoltan69:
@thenav56 I could not catch this exception in the https://github.com/IFRCGo/go-api/blob/develop/main/exception_handler.py#L14 or even in other method in this file.
The only success was to catch exception around https://github.com/IFRCGo/go-api/blob/develop/lang/admin.py#L92, but I think that is not a nice solution (and did not give fine result, just raised TransactionManagementError(
django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic')

thenav56:
@szabozoltan69 

Can you check what is the current response when using a DRF endpoint with **django_read_only** enabled?

DRF should catch and call our [custom exception handler](https://github.com/IFRCGo/go-api/blob/develop/main/exception_handler.py#L14), from where we should be able to check if `exc` is a [DjangoReadOnlyError](https://github.com/adamchainz/django-read-only/blob/main/src/django_read_only/__init__.py#L61) exception and then handle it from there.

While handling it we should do something like this https://github.com/IFRCGo/go-api/blob/develop/main/exception_handler.py#L29-L33 (without sentry as it is expected behavior)

Let me know if this is not the current behavior.

szabozoltan69:
The difficulty is that in case DjangoReadOnlyError can not be catched in custom_exception_handler. At least not by a pdb breakpoint.
If I generate a DjangoReadOnlyError, the result is this: https://porgeto.hu/tmp/ro_err.txt
kép

The really help would be to try it yourself and give a code hint, @thenav56

batpad:
Ah, I think the confusion here is errors returned by the API vs errors returned in the Admin. I think the 500 errors that Zoltan is seeing are coming from the Admin.

And @szabozoltan69 the custom exception handler will only be called if attempting to make Write calls on API endpoints. For the Admin, probably the easiest might be to just add a banner to the top when in Read-Only in bright red or something informing users that they will be unable to make any write operations currently?

thenav56:
Sorry, yeah I was referring to this
> Maybe a better feedback could be done via frontend. In case of write attempt: something like "hey, we are in maintenance mode, come back a bit later – site is in read only mode".

We can do something like this.
```diff
diff --git a/main/exception_handler.py b/main/exception_handler.py
index c0ae52e2..6f49ed09 100644
--- a/main/exception_handler.py
+++ b/main/exception_handler.py
@@ -3,6 +3,7 @@ import sentry_sdk
 from rest_framework.views import exception_handler
 from rest_framework.response import Response
 from rest_framework import status
+from django_read_only import DjangoReadOnlyError
 
 
 standard_error_string = (
@@ -17,20 +18,29 @@ def custom_exception_handler(exc, context):
 
     # For 500 errors, we create new response and add extra attributes to sentry
     if not response:
-        request = context.get('request')
-        if request and request.user and request.user.id:
-            with sentry_sdk.configure_scope() as scope:
-                scope.user = {
-                    'id': request.user.id,
-                    'email': request.user.email,
-                }
-                scope.set_extra('is_superuser', request.user.is_superuser)
-        sentry_sdk.capture_exception()
-        response_data = {
-            'errors': {
-                'non_field_errors': [standard_error_string]
-            },
-        }
+        # Expected ReadOnlyError
+        if type(exc) == DjangoReadOnlyError:
+            response_data = {
+                'errors': {
+                    'non_field_errors': ['Hey, we are in maintenance mode, come back a bit later – site is in read only mode']
+                },
+            }
+        else:
+            # Other 500 errors
+            request = context.get('request')
+            if request and request.user and request.user.id:
+                with sentry_sdk.configure_scope() as scope:
+                    scope.user = {
+                        'id': request.user.id,
+                        'email': request.user.email,
+                    }
+                    scope.set_extra('is_superuser', request.user.is_superuser)
+            sentry_sdk.capture_exception()
+            response_data = {
+                'errors': {
+                    'non_field_errors': [standard_error_string]
+                },
+            }
         response = Response(response_data, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
     return response
```

We will have this response on the API. Which should be handled automatically in the frontend (Where the new form library is used)
```json
{"errors":{"non_field_errors":["'Hey, we are in maintenance mode, come back a bit later – site is in read only mode"]}}
```

szabozoltan69:
Thanks for you both!
I used the code you shared, @thenav56 – it's working as we waited for frontend API requests.
Nevertheless the frontend login can not handle this kind of error, it shows simply:
image
The response was error 500:
{"errors":{"non_field_errors":["We are in maintenance mode, come back a bit later – site is in read only mode"]}}
Neither field report editing gave nice error feedback:

The copied error did not show the maintenance mode, but only this:
Request rejected by the server","responseText":"{\"detail\":\"Could not update field report\"}"}
If I tried to change my profile data, also a simple red "Error" appeared, but in Inspector the correct error message could be seen.
Maybe frontend login/fieldReport error handling should be fixed (via another ticket).

Now I look around about the Maintenance banner for Admin page.

szabozoltan69:
Tried not to hide anything behind the banner:
image

thenav56 (approved):
LGTM

PR #1725, opened 2023-03-28
Terraform + Helm

geohacker:
WIP


szabozoltan69 (commented):
Great check!

deploy/terraform/main.tf
```diff
@@ -61,7 +61,7 @@ module "resources" {
 
 terraform {
   backend "azurerm" {
-    resource_group_name  = "ifrcpgoterraform001rg"
+    resource_group_name  = "ifrctgo002rg"
```
geohacker, line 64: What was this change for @szabozoltan69? cc @batpad

geohacker: Has the storage account been migrated to the new resource group? Otherwise terraform may try to create new resources.

deploy/terraform/resources/variables.tf
```diff
@@ -190,6 +190,26 @@ variable "DEBUG_EMAIL" {
   default = "sanjay@developmentseed.org"
 }
 
+variable "RESOURCES_RG" {
```
geohacker, line 193: This is an extremely unnecessary change @szabozoltan69. We will have to revert all this. Sorry. 

@batpad

geohacker: We need these names visible as part of the templates because that's easy to debug. These resources are also managed manually and is a best practice to have these cleanly identified in the variables.tf. Any changes to these names should require a code update — that is intentional.

deploy/terraform/resources/variables.tf
```diff
@@ -283,4 +332,15 @@ locals {
   # prefixnodashes        = "${local.stack_id}${var.environment}"
   storage               = "${local.prefix}"
   deploy_secrets_prefix = "${local.prefix}"
+
+### Should work environment-wise.
+# ifrcgo_test_resources_rg = var.RESOURCES_RG    # Does not work this way.
+# ifrcgo_prod_resources_rg = var.RESOURCES_RG    # Does not work this way.
+# ifrcgo_test_resources_acr = var.RESOURCES_ACR  # Does not work this way.
+# ifrcgo_prod_resources_acr = var.RESOURCES_ACR  # Does not work this way.
+  ifrcgo_test_resources_db_server = var.RESOURCES_DB_SERVER
```
geohacker, line 341: 😭

szabozoltan69:
My only aim is to have as few wired-in constants as possible. Mainly the database entitiy names should not be in the open code. But anything else is also ugly wired in, which can be changed later.
Now I am away for some days. Do what you see the best.

deploy/terraform/main.tf
szabozoltan69: Yes, storage account and also acr was migrated to that resource group (which is not new though).

batpad:
Before we merge:

 - Make sure we change the branches things get deployed from to `develop` and `master`. @geohacker could you help with looking at this?

Issue #1746, opened 2023-04-12
(2023.04.12) is_deprecated to admin2 areas


szabozoltan69:
Similarly to [deployments_sector](https://dsgocdnapi.azureedge.net/static/database/tables/deployments_sector.html) and [api_district](https://dsgocdnapi.azureedge.net/static/database/tables/api_district.html), we should insert an **is_deprecated** boolean flag to [api_admin2](https://dsgocdnapi.azureedge.net/static/database/tables/api_admin2.html) table.

The backend (or frontend?) should handle the deprecated records.
(The same is a task for sectors and sectortags, if not done yet.)

We should enable the admin1- and admin2-updater jobs, so – when we want it – an update for a territory should create new (non-deprecated) records, while the old ones should become deprecated.

szabozoltan69:
@batpad @geohacker @davidmuchatiza

szabozoltan69:
Deployed to Staging.
(Unit tests still to be written)

szabozoltan69:
@batpad – as I promised, unit tests: https://github.com/IFRCGo/go-api/pull/1748/files

PR #1747, opened 2023-04-12
Admin2 area can be deprecated

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1746


batpad:
@szabozoltan69 this looks really good to me!

My only request would be to have some tests for this. It can be some pretty simple tests to just add one deprecated and non-deprecated Country, admin1 and admin2, and then simple tests to test all the cases and we are getting back what is expected. This is the kind of feature where any regression can be really problematic and it's just a good feeling to have tests for it.

szabozoltan69:
@batpad Good point!
I tested it via switching is_deprecated flags in district and country (and admin2), and checked the generated SQL queries also. They are ok.
I tested these:
```
curl http://localhost:8000/api/v2/admin2/?admin1__country__iso3=TUR
curl http://localhost:8000/api/v2/admin2/?admin1__country=174
curl http://localhost:8000/api/v2/district/?country__iso3=TUR
curl http://localhost:8000/api/v2/admin2/?admin1=3075
while:
update api_admin2 set is_deprecated =false where id=3075;
update api_admin2 set is_deprecated =false where id=77;
update api_admin2 set is_deprecated =true where id=77;
update api_country set is_deprecated=false where id=174;
update api_country set is_deprecated=true where id=174;
update api_district set is_deprecated=false where id=3075;
update api_district set is_deprecated=true where id=3075;
update api_district set is_deprecated=false where id=3075;

szabozoltan69:
Now I realized you meant some wired-in unit tests... Ok... let me see
 :-)

tovari:
@davidmuchatiza, would you mind to do some thorough tests on Staging? I suggest to select a country where we need to change  admin1 areas and create a FR (and emergency), 3W project, 3W activity related to the emergency page and a DREF, all covering the legacy admin1 areas. 
Then lets add the new admin1s, with new IDs, indicate deprecated the legacy admins. 
Then we should check, if the old admins are still available for the created content.
cc: @szabozoltan69

szabozoltan69:
Still not deployed yet...
The old admin2 areas are meant to be available only from backend Admin – they are not sent to frontend via API (by the plan).

batpad:
> Now I realized you meant some wired-in unit tests... Ok... let me see

Haa, sorry - I should have been more specific - yes, I meant adding to the unit tests :-)

szabozoltan69:
Deployed to Staging.

PR #1748, opened 2023-04-14
Districts and Admin2 areas can be deprecated – unit tests

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1746

### DIFF
--- a/api/test_views.py
+++ b/api/test_views.py
@@ -333,8 +333,8 @@ class HistoricalEventTest(APITestCase):
     def test_historical_events(self):
         region1 = models.Region.objects.create(name=1)
         region2 = models.Region.objects.create(name=2)
-        country1 = models.Country.objects.create(name='Nepal', iso3='nlp', region=region1)
-        country2 = models.Country.objects.create(name='India', iso3='ind', region=region2)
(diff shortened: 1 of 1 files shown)


api/test_views.py
```diff
@@ -418,3 +418,151 @@ def test_admin2_api(self):
...
+        response = self.client.get('/api/v2/district/').json()
+        self.assertEqual(response['count'], 2)
+
+        admin1_2.is_deprecated = True
+        admin1_2.save()
+        response = self.client.get('/api/v2/district/').json()
+        # one district deprecated, 2-1 = 1
+        self.assertEqual(response['count'], 1)
+
+        # Tear down
```
batpad, line 536: Am not sure if this `Tear down` comment makes sense in these places? Is that intentional?

szabozoltan69: Yes. My intention was to show: with that code part I "make back" the disturbed situation to a "nothing is deprecated" situation. Shall I use other expression for that?

batpad: Ah I see, makes sense. That's fine - as long as it was intentional, that is fine.

PR #1753, opened 2023-04-19
New appealdoc ingestor - wip

szabozoltan69:
API usage instead of scraping.
Refers to https://github.com/IFRCGo/go-api/issues/1754


szabozoltan69:
Refers also to https://github.com/IFRCGo/go-api/issues/1755

Issue #1460, opened 2022-06-21
Implement generalized cache for API responses


batpad:
We should setup a caching layer for API responses that we can tweak per endpoint based on needs.

Currently, ever API request hits the database, and for a lot of these requests, we can cache responses for some time, to reduce load on the database and drastically improve site performance.

Django has a relatively decent cache framework that we can use, but I'd be super interested in seeing if this can be implemented at the CDN level without having to touch the application code. (cc @MartonNagyATOS )

Let's discuss approaches here and hopefully are able to implement something soon and then iterate.

cc @szabozoltan69 @tovari @frozenhelium

MartonNagyATOS:
Hi All,

Without knowing many details, in general it is possible to tunnel the API calls through CDN.
CDN is not just caching the data, but deployed to edge servers, meaning that the client calls will be served by an edge server that is closest to client. (lower latency and roundtrip time)
If one of the edge servers fails, it automatically switches to the next closest location, so it also increases the availability for the cached responses.
And there are many other options at the CDN level, like response compression to reduce the traffic if the user agent supports it. (All modern browser support it.)

Marton

From: Sanjay Bhangar ***@***.***>
Sent: Tuesday, 21 June 2022 13:24
To: IFRCGo/go-api ***@***.***>
Cc: Nagy, Marton ***@***.***>; Mention ***@***.***>
Subject: [IFRCGo/go-api] Implement generalized cache for API responses (Issue #1460)


Caution! External email. Do not open attachments or click links, unless this email comes from a known sender and you know the content is safe.

We should setup a caching layer for API responses that we can tweak per endpoint based on needs.

Currently, ever API request hits the database, and for a lot of these requests, we can cache responses for some time, to reduce load on the database and drastically improve site performance.

[...]

PR #1758, opened 2023-04-24
Setup caching

oliverroick:
Addresses #1460

Setting up caching for HTTP responses

## Changes

* Configuration of Django's Cache Framework, using Redis as the caching backend. Cached entries are written to Redis DB `1` to separate the cached entries from the Celery database.
* Pages are cached for 10 minutes by default
* The time can be adapted for each view individually (see the `list` method of `AppealViewset` as an example)
* Add custom cache middlewares, which sets a per-user cache-key prefix, which allows separate caches for each user

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


oliverroick:
@batpad The tests are now passing except for the one snapshot test

szabozoltan69:
You could try to get deployments/snapshots/snap_tests.py from develop branch, it could be modified since you forked from develop.

batpad:
@oliverroick this looks great !

@szabozoltan69 let's discuss this on our next developer call - essentially, this setup will by default cache all API endpoints for 10 minutes. This is probably not what we want across all the end-points, so we will likely need to do some further work here before merging:

 - Figure out which endpoints we should cache by default for how long: some endpoints we may want no cache at all, some a small cache is fine, etc. We can discuss this with @tovari .
 - Figure out if we need to implement cache invalidation for some endpoints - i.e. make sure we bust the cache when something about that model is changed - this will be a bit of work and we should do this on a case-by-case basis when we know we want to bust the cache under specific circumstances.
 - Figure out which endpoints we display different data to the user based on permission level. We will need to ensure the user key is accounted for in the caching - cc @oliverroick 

My general sense is that we will get huge benefits even implementing a cache with a relatively small timeout value - eg. 2 minutes or so - we immediately will greatly reduce the load on the database, especially for queries made by every user to the site eg. the home-page. Let's discuss the details and trade-offs on the developer call.

batpad:
Thanks @oliverroick this is looking good to me!

@thenav56 do you think you'd be able to do a review here?

If things look good to you in terms of the code and setup, then we can have another discussion around the logic during the dev call and see if we need to make changes per business requirements.

api/test_cache.py
```diff
@@ -0,0 +1,127 @@
...
+FAKE_REDIS_CACHE = {
+    'default': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': env('CACHE_REDIS_URL'),
+        'OPTIONS': {
+            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
+            'connection_class': FakeConnection
+        }
+    }
+}
```
thenav56, line 19: It should be fine to use a Memory cache for testing as we aren't using any Redis-specific features.
```python
    'default': {
        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
        'LOCATION': 'unique-snowflake-rest',
    }
```

Or we can use a different database index for testing (max supported is 16 by default)
For eg.
Define env variables for testing for docker-compose and then use them in settings.py and here
```bash
CELERY_REDIS_URL: ${CELERY_REDIS_URL:-redis://redis:6379/0}
CELERY_TEST_REDIS_URL: ${CELERY_TEST_REDIS_URL:-redis://redis:6379/10}
CACHE_REDIS_URL: ${CACHE_REDIS_URL:-redis://redis:6379/1}
CACHE_TEST_REDIS_URL: ${CACHE_TEST_REDIS_URL:-redis://redis:6379/11}
```
And  make TEST_* as optional in settings.py

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
-DJANGO_SETTINGS_MODULE = main.settings
+DJANGO_SETTINGS_MODULE = main.settings-test
```
thenav56, line 2:
```suggestion
DJANGO_SETTINGS_MODULE = main.settings_test
```

main/settings-test.py
thenav56, line 1: We were using override_settings in base test classes defined in main/test_case.py.
Maybe move those here as well.
https://github.com/IFRCGo/go-api/blob/develop/main/test_case.py (Search for override_settings)

thenav56 (changes requested):
Looks good. Few comments

main/settings-test.py
oliverroick: Thanks for the suggestion. I've removed the test-settings file in used the `override_settings` decorator to specify the cache for tests.

api/test_cache.py
oliverroick: These tests don't actually use Redis. I'm using the fakeredis library, hence the `FakeConnection` connection class. But I can use Redis instead if that's preferred. 

I decided against `LocMemCache` to keep the test environment close to production. We're not using anything Redis-specific now but we might in the future.

thenav56: Then, using Redis directly makes sense here.
Want to avoid adding a new library that simulates a service that we already have running :D

oliverroick: Ok. I changed that so the cache tests use Redis and removed the dependency

thenav56 (approved):

szabozoltan69:
Merged and deployed to Staging.

szabozoltan69:
@thenav56 we should set value for CACHE_REDIS_URL, otherwise we get a KeyError: 'CACHE_REDIS_URL'
Otherwise the system gets down (which happened on Staging).
@oliverroick could you please collect the new env variables we should set before deployment?

thenav56:
@szabozoltan69 Yes we need to set the value. You can use another index for CACHE similar to CELERY Redis env.
For eg:
```
IF CELERY_REDIS_URL=redis://redis:6379/1
then CACHE_REDIS_URL=redis://redis:6379/2 works
```

szabozoltan69:
We have CELERY_REDIS_URL=redis://redis:6379/0 so maybe I use a /1 for that.

szabozoltan69:
@batpad @geohacker please note that a new variable arrived to the backend deployment system. (We should have a forum to administer these better.)

szabozoltan69:
@thenav56 now the new system works fine on Staging

oliverroick:
@szabozoltan69  If you want to generally switch off caching, then I just wouldn't load the middleware. For individual views you use the [never_cache](https://docs.djangoproject.com/en/3.2/topics/http/decorators/#django.views.decorators.cache.never_cache)

szabozoltan69:
@oliverroick You mean that commenting out https://github.com/IFRCGo/go-api/blob/develop/main/settings.py#L217 and https://github.com/IFRCGo/go-api/blob/develop/main/settings.py#L219 (middlewares.cache.UpdateCacheForUserMiddleware and middlewares.cache.FetchFromCacheForUserMiddleware) is enough and will not cause any runtime errors?
The aim of "not using (still now)" is to leave a bit of time for us to think about better strategies (than the 10 minutes limit), but we would be able to release the develop branch to production.

oliverroick:
> You mean that commenting out https://github.com/IFRCGo/go-api/blob/develop/main/settings.py#L217 and https://github.com/IFRCGo/go-api/blob/develop/main/settings.py#L219 (middlewares.cache.UpdateCacheForUserMiddleware and middlewares.cache.FetchFromCacheForUserMiddleware) is enough and will not cause any runtime errors?

yes correct, the middleware is what controls the caching. the rest are just settings and Django's cache middleware puts the cached data in the right place based on those. 

What I would check, though, is that [this line](https://github.com/IFRCGo/go-api/blob/develop/api/drf_views.py#L652) does cause any troubles when the caching middleware isn't configured. 

For the tests: Yes, you'd have to deactivate them, or configure the test settings to include the cache for these tests only

szabozoltan69:
(Proposed changes are done: https://github.com/IFRCGo/go-api/commit/9c2f0de751cced40f0c466a33de67854d39f3c27 )

Issue #1755, opened 2023-04-20
3W project - districts should be optional


tovari:
We aim to standardize, how users can report geo coverage for an entire country. At the moment, only at 3W projects, all admin1 have to be selected, while at other reports, admin selection is optional.

Please make the district entry optional at 3W projects as well. On the frontend UI as well as for the API import.

szabozoltan69:
Dear @frozenhelium – please give me some hints, how can I remove the frontend "district is mandatory" requirement in 
http://localhost:3000/three-w/2499/edit/
and
http://localhost:3000/emergency-three-w/127/edit/
pages. (If you have no time, I think I will find it myself also, but maybe it will be faster and safer with your hint.)

szabozoltan69:
Deployed to Staging.
You can try: no districts in https://go-stage.ifrc.org/three-w/20/ and in https://go-stage.ifrc.org/emergency-three-w/20/
We should test that https://go-stage.ifrc.org/emergency/4422/#activities page is OK (mainly the map), where we are redirected after editing the above emergency-three-w/20/ .

szabozoltan69:
Now the Country 3W maps are ok also:
![2023-04-27_10-40-25](https://user-images.githubusercontent.com/6050643/234813225-397a5a38-7e23-46eb-9d49-363f88d1c561.png)

![2023-04-27_10-43-42](https://user-images.githubusercontent.com/6050643/234813252-2fb214a7-8fad-4f62-8b78-c38ff53884e2.png)

With multiple projects:
![2023-04-27_10-43-49](https://user-images.githubusercontent.com/6050643/234813272-15fc6e93-5d26-4fec-8867-a7326ba66249.png)

szabozoltan69:
Deployed to Staging.

szabozoltan69:
Project import can be tested from now on Staging Admin, all (known) bugs were fixed. – @tovari

PR #1763, opened 2023-04-25
Updating appeal_documents. ISO should be unique.

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1755


bewakes:
@szabozoltan69  There is a problem loading Districts fixture: `Country id 229 does not exist.` I checked both the fixture files and Districts has a district referencing country 229 while Countries.json does not have a country with id 229.

szabozoltan69:
The reason is that I dopped the disputable and deprecated areas from countries. Feel free to drop those districts also. Or I can fix them too in a few hours.

PR #1778, opened 2023-05-09
Production stack

geohacker:
WIP


deploy/terraform/resources/variables.tf
```diff
@@ -202,6 +203,28 @@ variable "ifrcgo_test_resources_db" {
...
+}
+
+variable "ifrcgo_prod_resources_rg" {
+  type = string
+  default = "ifrcpgo002rg"
+}
+
+variable "ifrcgo_prod_resources_db_server" {
+  type = string
+  default = "ifrcpgokdb001"
```
szabozoltan69, line 219: Is it just a default string which can/will be overridden from an ENV variable? Or (if we move to another named database) should we change the code here?

PR #1791, opened 2023-05-30
fix admin redirect

batpad:
Change the redirect logic so that if a user visits https://goadmin.ifrc.org it will still redirect to https://goadmin.ifrc.org/admin, but for other 404 URLs, it will correctly return 404s.

cc @tnagorra @thenav56 @szabozoltan69 @k9845


szabozoltan69:
Perfect solution with backward compatibility!
We should still change on frontend "Edit Event" and "Edit Country".

Issue #1794, opened 2023-06-02
[PROD] EmergencyProjectActivity history to EmergencyProjects


szabozoltan69:
## Issue
If an activity item of an emergency project changes, it can not be seen in the emergency project history.
Would be needed!

szabozoltan69:
Deployed to Prod and Staging.

PR #1796, opened 2023-06-02
Follow project in revision

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1794


szabozoltan69:
It is not clear, why I can't see the "compare" buttons, but just some rows about the changes:
kép

While in other similar parts (e.g. for AnnualSplits) there is "Compare":
kép
@thenav56 @batpad

PR #1811, opened 2023-06-13
Switch on cache

szabozoltan69:
Cache for anonymous users only


main/settings.py
```diff
@@ -206,21 +206,21 @@
 }
 
 MIDDLEWARE = [
+    'middlewares.cache.UpdateCacheForUserMiddleware',
```
thenav56, line 209: Since we are using `django.contrib.auth.middleware.AuthenticationMiddleware` in this middlware. We will need to move this after that.

main/settings.py
```diff
@@ -206,21 +206,21 @@
...
     'django.contrib.auth.middleware.AuthenticationMiddleware',
-    # ¤ 'middlewares.cache.UpdateCacheForUserMiddleware',
     'django.middleware.common.CommonMiddleware',
-    # ¤ 'middlewares.cache.FetchFromCacheForUserMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'middlewares.middlewares.RequestMiddleware',
     'reversion.middleware.RevisionMiddleware',
+    'middlewares.cache.FetchFromCacheForUserMiddleware',
```
thenav56, line 223: We will need to move this back as well, as we will need headers from other middlewares as well.

thenav56 (changes requested):

api/drf_views.py
```diff
@@ -14,8 +14,8 @@
 from django.db.models import Prefetch, Count, Q, OuterRef
 from django.utils import timezone
 
-# ¤ from django.utils.decorators import method_decorator
-# ¤ from django.views.decorators.cache import cache_page
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_page
```
thenav56, line 18: Let's remove unused imports as well.

thenav56 (changes requested):

thenav56 (approved):

Issue #1804, opened 2023-06-07
[PROD] TinyMCE – rich text editor, pasted images


szabozoltan69:
## Issue
Instead of Base64 format "pasted images" we should solve real image file uploads to our server.
Based on this: https://www.tiny.cloud/docs/general-configuration-guide/upload-images/
This can be also helpful: https://dev.to/zankoan/upload-local-imagevideo-in-django-tinymce-text-editor-mkn

Local images that are added through other means are also uploaded using this function, such as images added by drag and drop when using the [paste_data_images](https://www.tiny.cloud/docs/plugins/opensource/paste/#paste_data_images) configuration property, or using the Tiny [PowerPaste plugin](https://www.tiny.cloud/docs/plugins/premium/powerpaste/).

Unfortunately:
https://www.tiny.cloud/docs/plugins/premium/powerpaste/
This plugin is only available for paid TinyMCE subscriptions. (min 37.5 USD / month)
But via this: https://gist.github.com/AshishPandagre/c6a3b0f5ff4724ed5d29b4ff347828f4 maybe we can do it either.

szabozoltan69:
Dear @thenav56 @batpad @k9845 –
I have some progress in this topic.
My issue is: this https://gist.github.com/AshishPandagre/c6a3b0f5ff4724ed5d29b4ff347828f4#file-app_views-py - row 7, csfr_exempt seems to be not working.
For local test I can make a hack via:
cd /usr/local/lib/python3.8/site-packages/rest_framework; vi authentication.py
And putting a return to    def enforce_csrf(self, request), but for a real solution we should be able to define a better short-circuit of CSRF for this image upload function.

This way I have some images which were upload via copy-and-paste:
```
root@d166c262ed84:~/go-api/media/tinymce# ls -l
-rw-r--r--  1 root root  627 Jun 21 08:31 blobid0.png
-rw-r--r--  1 root root  561 Jun 21 08:37 blobid1.png
-rw-r--r--  1 root root  808 Jun 21 09:05 blobid2.png
-rw-r--r--  1 root root 2117 Jun 21 09:06 blobid3.png
```
Please check [the PR ](https://github.com/IFRCGo/go-api/pull/1828) if you have time.

szabozoltan69:
@batpad @thenav56 if you have any idea about this ^ – I would highly appreciate it.

batpad:
[...]

PR #1828, opened 2023-06-21
Automatic image upload from rich text editor

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1804


batpad:
@thenav56 could you take a look?

I am guessing we probably want to upload the images to Azure Blob Storage or so, rather than just on the local file-system? Otherwise, broadly this looks good and like what we need to do here.

api/views.py
```diff
@@ -50,6 +53,45 @@ def unauthorized(message="You must be logged in"):
...
+
+        with open(file_path, 'wb+') as f:
+            for chunk in file_obj.chunks():
+                f.write(chunk)
+
+        return JsonResponse({
+            'message': 'Image uploaded successfully',
+            'location': file_url
+        })
+    return JsonResponse({'detail': "Wrong request"})
```
thenav56, line 92: Hi @szabozoltan69

For uploading, we can directly use the storage class instead.
Something like this
```python3
from django.core.files.storage import FileSystemStorage, get_storage_class

DefaultMediaStorageClass = get_storage_class()


def get_full_media_url(media_path):
    if DefaultMediaStorageClass == FileSystemStorage:
        return f"{settings.HTTP_PROTOCOL}://{settings.API_FQDN}{media_path}"
    # For other storage. Assuming they provide absolute URL
    return media_path


def func...():
    ...
    storage = DefaultMediaStorageClass()
    storage.save(file_path, content_file)
    file_url = get_full_media_url(file_path)
```

> WHERE `get_storage_class` provides storage defined using `DEFAULT_FILE_STORAGE`
https://docs.djangoproject.com/en/3.2/ref/files/storage/#django.core.files.storage.get_storage_class

main/settings.py
```diff
@@ -282,6 +282,7 @@
 ]
 
 TINYMCE_DEFAULT_CONFIG = {
+    'images_upload_url': '/upload_image',  # Image upload address in Django route
```
thenav56, line 285: Maybe we can use reverse_lazy.
https://docs.djangoproject.com/en/3.2/ref/urlresolvers/#reverse-lazy

api/views.py
```diff
@@ -50,6 +53,45 @@ def unauthorized(message="You must be logged in"):
     return JsonResponse({"statusCode": 401, "error_message": message}, status=401)
 
 
+@csrf_exempt
+def upload_image(request):
```
thenav56, line 57: - Restrict with logged-in users only
- Track these uploads (Suggestion)
  - Maybe create a table with a columns file path, created by, created at

thenav56 (changes requested):

api/views.py
thenav56: Maybe we can use images_upload_handler to provide CSRF+Auth
https://www.tiny.cloud/docs/demo/local-upload/#

api/views.py
szabozoltan69: Thanks @thenav56

I wonder how content_file variable should have it's content.

thenav56: I get confused with this, but something like this should do :D 
```python
from django.core.files.base import ContentFile

ContentFile(
        file.read().encode('utf-8')
    )

```

szabozoltan69: Well, I could not use this file.read (not obvious which would be the "file"), but maybe file_obj.chunks() could be used like:
```
        with open(file_url, 'wb+') as f:
            for chunk in file_obj.chunks():
                f.write(chunk)
```

thenav56: Maybe `ContentFile(file_obj)` works?
Or directly `storage.save(file_path, file_obj)`

main/settings.py
szabozoltan69: I was not able to get this reverse_lazy work. Please give me some more hints.
I gave a name in urls.py, but there are other tasks also.
```
    url(r"^upload_image", upload_image, name="upload_image"),
```

szabozoltan69:
The "rest_framework.exceptions.PermissionDenied: CSRF Failed: CSRF token missing or incorrect" should be fixed.

main/settings.py
thenav56: Hey @szabozoltan69 Did you mean there are other URLs with the same name? Maybe we can use a different name instead? Like "tinymc_upload_image"
Let me know if I understood incorrectly.

szabozoltan69: Hey @thenav56 , there is no other same name, just for the sake of reverse_lazy I gave a name to the endpoint. But I could not do the next steps about it. Maybe not even meeded to be named.

szabozoltan69: As "other tasks" I meant other developing tasks for sake of reverse_lazy.

thenav56: Okay. Let's leave with a TODO there then.

szabozoltan69:
@batpad @thenav56 Please have a look at this, mainly the CSRF "protection" is questionable.

tnagorra:
@szabozoltan69 Aren't we disabling "image upload from rich text editors" for the time being?

cc: @batpad @frozenhelium @thenav56

szabozoltan69:
Yes, we disabled "image paste from rich text editors". The image link placing is still ok from Admin page (and I think it should be also from the client side), but let's close this ticket due to the "automation" is not on our plate.

PR #1829, opened 2023-06-22
Add openapi endpoint

thenav56:
## Changes
Add base open API v3 endpoint
- Using [drf-spectacular](https://drf-spectacular.readthedocs.io/en/latest/) for schema generation
- Endpoint:
  - http://localhost:8000/api-docs/ -> YAML
  - http://localhost:8000/api-docs/swagger-ui/
  - http://localhost:8000/api-docs/redoc/
- Replace existing /docs/ with redoc ui
  
> NOTE: Make sure to build a docker image

```bash
docker-compose exec serve ./manage.py spectacular --color
docker-compose exec serve ./manage.py spectacular --color --file openapi-v3.yml
```
```
Schema generation summary:
Warnings: 109 (89 unique)
Errors:   224 (50 unique)
```

### DIFF
--- a/main/urls.py
+++ b/main/urls.py
@@ -73,8 +73,6 @@ from local_units.views import LocalUnitListAPIView, LocalUnitDetailAPIView
 
 # DRF routes
 from rest_framework import routers
-from rest_framework.documentation import include_docs_urls
-from rest_framework import permissions
 from drf_spectacular.views import SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView
 from api import drf_views as api_views
 from flash_update import views as flash_views
@@ -240,11 +238,9 @@ urlpatterns = [
     url(r"^exception-error-for-devs", DummyExceptionError.as_view()),
     path("i18n/", include("django.conf.urls.i18n")),
     # Docs
-    url(r"^docs/", include_docs_urls(title="IFRC GO API", public=False)),  # TODO: Remove this?
+    path("docs/", SpectacularRedocView.as_view(url_name='schema'), name='redoc'),
     path("api-docs/", SpectacularAPIView.as_view(), name='schema'),
     path("api-docs/swagger-ui/", SpectacularSwaggerView.as_view(url_name='schema'), name='swagger-ui'),
-    path("api-docs/redoc/", SpectacularRedocView.as_view(url_name='schema'), name='redoc'),
-
 ]
 
 if settings.DEBUG:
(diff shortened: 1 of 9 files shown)


main/urls.py
```diff
@@ -168,6 +171,17 @@
 
 router.register(r"review-country", api_views.CountryOfFieldReportToReviewViewset, basename="review_country")
 
+api_schema_view = get_schema_view(
+    openapi.Info(
+        title="DEEP API",
+        default_version='v1',
+        description="DEEP API",
+        contact=openapi.Contact(email="admin@thedeep.io"),
```
frozenhelium, line 179: Let's update these info for IFRC

frozenhelium (changes requested):

main/urls.py
thenav56: Done

tnagorra:
## OpenAPI v3

We have updated the OpenAPI to version 3.
It includes a command to show all of the errors/warning related to schema

Here is an output of `docker-compose exec serve ./manage.py spectacular --color`

image

## Redoc

We are now using Redoc to generate documentation for http://localhost:8000/docs/
There might be minor differences from the previous generated documentation.

## Next Steps

- Fix the errors/warning listed gradually.
- Check if the generated types are correct for all of the endpoints.

@batpad @szabozoltan69 Can we also discuss on how we can continue this work and improve the generated OpenAPI types.

frozenhelium:
@szabozoltan69 please review and merge this

frozenhelium (approved):

szabozoltan69 (approved):

batpad:
Spectacular work, @thenav56 @tnagorra ! Let's discuss how we can fix up errors on the dev call or so.

szabozoltan69:
Merged to develop and deployed to Staging. (Error fixing later)

szabozoltan69:
kép

szabozoltan69:
@thenav56 @tnagorra is it possible to use this /docs endpoint interactively? So, giving some arguments and sending the request?

szabozoltan69:
Or maybe could we keep the old possibility as a /docs_v1 ?

thenav56:
@szabozoltan69 swagger-ui has interaction, maybe we can use that http://localhost:8000/api-docs/swagger-ui/

tnagorra:
@szabozoltan69 I also suggest we use https://goadmin-stage.ifrc.org/api-docs/swagger-ui/ for interactivity.

szabozoltan69:
I am not sure that this kind of output structure could remain, which is in https://goadmin-stage.ifrc.org/api/v2/brief
It's just a healthcheck endpoint, for human reading, not for any other machine usage.
(It's the only one which is "hacked" like this.)

thenav56:
I think that should be okay. Basically, we define a serializer that we use to generate the data instead of a manual response. Then the schema will show the structure.

PR #1838, opened 2023-07-03
Fix Sentry issue 1

szabozoltan69:
Refers to Sentry issues/1/

### DIFF
--- a/registrations/views.py
+++ b/registrations/views.py
@@ -21,6 +21,8 @@ from .models import Pending, DomainWhitelist
 
 
 def is_valid_domain(email):
+    if email is None:
+        return False
     domain = email.lower().split('@')[1]
     is_allowed = (
         DomainWhitelist.objects


registrations/views.py
```diff
@@ -21,6 +21,8 @@
 
 
 def is_valid_domain(email):
+    if email is None:
```
thenav56, line 24: Maybe we can also make the email required field?

thenav56 (changes requested):

registrations/views.py
szabozoltan69: On frontend it is a required field. (And also here, on backend, it is not optional.)
I suppose some non-frontend request can send a NoneType argument.
Or did I misunderstand something? (In my knowledge an optional argument is like something=defaultvalue, and if we do not give a defaultvalue, it is a mandatory = required argument.)

szabozoltan69: @thenav56

thenav56: Looks like registration is handling data in a non-DRF way which directly looks at the requested data.
For now, this is fine, but we should track this as an issue to change the logic to use DRF serializer instead for better data validation.

https://github.com/IFRCGo/go-api/blob/ef25c1d298020f98b331da140def06ca60fc9fd7/registrations/views.py#L83-L118

thenav56 (approved):

Issue #1812, opened 2023-06-14
Build connection between surge alerts and deployments tables


JonathanGarro:
**Is your feature request related to a problem? Please describe.**

As discussed on a call on 2023-06-14 with Sanjay and Zoltan, in order to meet requirements from the Surge team, we need to build a way for end users to connect a given deployment to the surge alert that it is filling.

**Describe the solution you'd like**
I presume that when the RRMS team updates a deployment, that they are recording the surge alert that it is filling, but that data is not readily available in the API. From an end-user perspective, as long as we link the surge alert `id` to the deployment, that should suffice. 

**Should there be any restrictions around viewing or updating information related to this feature**
This data should only be available to users that pass an API key, given that it references personal information as part of the deployment.

anamariaescobar:
looping @alicefahey into this ticket

batpad:
The response for `deployments` from the Molnix API has a `position_id` field. This can be NULL, but when it exists, it refers to the Molnix ID for the Alert that this deployment is filling.

So, we will need to:

 - Add a NULLable ForeignKey from PersonnelDeployment to SurgeAlert in our models
 - Populate the foreign key by fetching the SurgeAlert with the `molnix_id` referenced by the `position_id` when it exists, when parsing through the deployments response from Molnx, so around here: https://github.com/IFRCGo/go-api/blob/develop/api/management/commands/sync_molnix.py#L224

PR #1854, opened 2023-07-26
Build connection between surge alerts and deployments

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1812


batpad:
Looks great to me!

batpad (approved):

PR #1862, opened 2023-08-02
Feature/dref regional admin

k9845:
Addresses
- https://github.com/IFRCGo/go-frontend/issues/2868
## Changes
 - Allow regional admin to view and edit dref

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release

### DIFF
--- a/dref/views.py
+++ b/dref/views.py
@@ -71,9 +71,10 @@ class DrefViewSet(RevisionMixin, viewsets.ModelViewSet):
                     dref_list.append(_queryset)
             drefs = []
             for dref in dref_list:
-                id = list(dref.values_list('id', flat=True))
-                new_dref = Dref.objects.filter(id__in=id).first()
-                drefs.append(new_dref.id)
+                ids = list(dref.values_list('id', flat=True))
+                for id in ids:
+                    new_dref = Dref.objects.filter(id=id).first()
+                    drefs.append(new_dref.id)
             if len(drefs):
                 return queryset.filter(models.Q(id__in=drefs) | models.Q(created_by=user)).distinct()
             else:
@@ -137,9 +138,10 @@ class DrefOperationalUpdateViewSet(RevisionMixin, viewsets.ModelViewSet):
                     dref_op_list.append(_queryset)
             ops = []
             for op in dref_op_list:
-                id = list(op.values_list('id', flat=True))
-                new_dref_op = DrefOperationalUpdate.objects.filter(id__in=id).first()
-                ops.append(new_dref_op.id)
+                ids = list(op.values_list('id', flat=True))
+                for id in ids:
+                    new_dref_op = DrefOperationalUpdate.objects.filter(id=id).first()
+                    ops.append(new_dref_op.id)
             if len(ops):
                 return queryset.filter(models.Q(id__in=ops) | models.Q(created_by=user)).distinct()
             else:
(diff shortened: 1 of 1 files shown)


dref/views.py
```diff
@@ -59,8 +60,25 @@ def get_queryset(self):
         )
         if user.is_superuser:
             return queryset
-        else:
-            return Dref.get_for(user)
+        elif not user.is_superuser:
+            # get current user dref regions
+            regions = [0, 1, 2, 3, 4]
```
thenav56, line 65: Why are we using raw values here?

dref/views.py
```diff
@@ -59,8 +60,25 @@ def get_queryset(self):
...
+            drefs = []
+            for dref in dref_list:
+                ids = list(dref.values_list('id', flat=True))
+                for id in ids:
+                    new_dref = Dref.objects.filter(id=id).first()
+                    drefs.append(new_dref.id)
+            if len(drefs):
+                return queryset.filter(models.Q(id__in=drefs) | models.Q(created_by=user)).distinct()
+            else:
+                return Dref.get_for(user)
```
thenav56, line 81: This has lots of N+1 queries, Can we do something like this instead?

```suggestion
            dref_admin_regions_id = [
                codename.replace('dref_region_admin_', '')
                for codename in Permission.objects.filter(
                    group__user=user,
                    codename__startswith='dref_region_admin_',
                    # Maybe add additional filter to make sure dref_region_admin_<number> are only returned
                ).values_list('codename', flat=True)
            ]
            if len(dref_admin_regions_id):
                return queryset.filter(
                    models.Q(created_by=user)
                    | models.Q(country__region__in=dref_admin_regions_id)
                ).distinct()
            else:
                return queryset.model.get_for(user)
```

thenav56 (changes requested):

dref/views.py
```diff
@@ -45,6 +46,23 @@
...
+                group__user=user,
+                codename__startswith='dref_region_admin_',
+            ).values_list('codename', flat=True)
+        ]
+        if len(dref_admin_regions_id):
+            return queryset.filter(
+                models.Q(created_by=user) | models.Q(country__region__in=dref_admin_regions_id)
+            ).distinct()
+        else:
+            return queryset.model.get_for(user)
```
thenav56, line 63: Let's move the is_superuser logic here as well.
```suggestion
    if user.is_superuser:
        return queryset
    # Check if regional admin
    dref_admin_regions_id = [
        codename.replace('dref_region_admin_', '')
        for codename in Permission.objects.filter(
            group__user=user,
            codename__startswith='dref_region_admin_',
        ).values_list('codename', flat=True)
    ]
    if len(dref_admin_regions_id):
        return queryset.filter(
            models.Q(created_by=user) | models.Q(country__region__in=dref_admin_regions_id)
        ).distinct()
    # Normal access
    return queryset.model.get_for(user)
```

dref/views.py
```diff
@@ -60,7 +78,8 @@ def get_queryset(self):
         if user.is_superuser:
             return queryset
         else:
-            return Dref.get_for(user)
+            return filter_dref_queryset_by_user_access(user, queryset)
```
thenav56, line 81:
```suggestion
        return filter_dref_queryset_by_user_access(user, queryset)
```

thenav56 (changes requested):
Looks good, last comment

thenav56 (approved):

Issue #1814, opened 2023-06-14
Add additional RRMS fields to surge alert and deployment tables


JonathanGarro:
As referenced on the 2023-06-14 call, there are additional fields we need to add to these tables in order to meet the surge team's requirements for a visualization on the GO Platform reflecting various KPIs and common queries. Most of the data we need is already available, save for a few additions: 

1. Funding: Type = str
2. Contract_Type: Type = str
3. Location: Type = str (this will refer to the city and country where the person deploys)
4. Gender: Type = str
5. Appraisal_Received = boolean

batpad:
1, 2, 3 should be for Alerts - 4,5 for Deployments.

Next steps: @batpad checks current API from Molnix and if these fields exist - then just need to add them to GO and make sure they are imported.

We can discuss backfill once we have code working to ingest current data.

anamariaescobar:
looping @alicefahey into this ticket

batpad:
@JonathanGarro I'm not seeing these fields in the API responses for either `deployments` or `open positions`. The only field of the above that I see is `contract_id` and `contract` in the response for Deployments, though I've not been able to find an example in the staging data where `contract` is not NULL.

I think next steps here:

 - If you can point out to me where in the Staging Molnix UI these things appear, I can do some inspecting and try and figure out what API endpoint they are coming from.
 - It might be useful to get on a call and explore the network requests and API together to figure out which endpoints to get this data from.
 - For `contract`, if we do want the contract information for Deployments, it would be useful to have an example with actual data to see what that looks like.

JonathanGarro:
Hey @batpad - Sure thing. In the staging site, when you look at the Deployments tab and drill down to a single deployment, the Funding and Contract fields appear:

![Screen Shot 2023-07-24 at 09 12 06 AM](https://github.com/IFRCGo/go-api/assets/8890661/5c8bbcbc-d353-423f-b471-8fabb8d63327)

For location, its a separate object that I'm guessing is associated via a deployment ID? Here's what it looks like on the front end: 

![Screen Shot 2023-07-24 at 09 13 09 AM](https://github.com/IFRCGo/go-api/assets/8890661/d3470e3b-8277-4063-a68e-f9d33b35a231)

Ditto the Appraisals - I think that might require a bit of filtering logic as it includes a few different attachment types, so we'd need to just set a boolean to `True` if `attachment_type` (or whatever it's called) `== 'Appraisal'`

[...]

PR #1867, opened 2023-08-04
Fetch appraisals score, gender and location from Molnix to Personnel

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1814


api/management/commands/sync_molnix.py
```diff
@@ -242,6 +266,14 @@ def sync_deployments(molnix_deployments, molnix_api, countries):
         personnel.country_to = country_to
         country_from = None
         personnel.surge_alert = surge_alert
+        personnel.appraisal_score = appraisal_score
+        if gender == 'male':
```
szabozoltan69, line 270: That part could be nicer.

batpad: @szabozoltan69 let's just keep this a CharField for now instead of enforcing a CHOICES

api/management/commands/sync_molnix.py
```diff
@@ -242,6 +266,14 @@ def sync_deployments(molnix_deployments, molnix_api, countries):
         personnel.country_to = country_to
         country_from = None
         personnel.surge_alert = surge_alert
+        personnel.appraisal_score = appraisal_score
```
batpad, line 269: @szabozoltan69 many apologies - after discussing with Jonathan, it was decided that we should not store `appraisal_score` but just an `appraisal_received` boolean field. Sorry about this.

So here, we can just do something like `personnel.appraisal_received = len(md['appraisals']) > 0`

szabozoltan69:
Hi @k9845 , could you please have a look at the below test failure?
I think you can fix it better than me:
FAILED dref/test_views.py::DrefTestCase::test_dref_is_published - AssertionError: 400 != 200 : b'{"modified_at":"OBSOLETE_PAYLOAD"}'

szabozoltan69:
Maybe these wired-in dates and intervals are the root cause:
```
        initial_now = datetime(2023, 7, 28)
        ...
        data["modified_at"] = initial_now + timedelta(days=10)
```

szabozoltan69:
Thanks @k9845

szabozoltan69:
@k9845 yes, I continue my work with git pull. Thx.

batpad (approved):

PR #1877, opened 2023-08-18
Facilitate setting up an additional domain for the api

sunu:
Changes to the infrastructure code to allow adding an additional domain for the API

## Changes

* Adds a new terraform variable called `additionalDomain`
* This gets populated from a env var called `API_ADDITIONAL_FQDN` in the azure pipeline

## Checklist 

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

@batpad @szabozoltan69 `PRODUCTION_ADDITIONAL_URL` secret needs to be set in Azure secrets to populate the additional url.


batpad:
@szabozoltan69 whenever you're around, we can set that secret and then merge this and test if this works.

szabozoltan69:
Deployed to terraform prod

PR #1886, opened 2023-08-24
add annotations to increase timeouts for nginx ingress

batpad:
cc @szabozoltan69 - we should deploy this and check if it improves the timeout situations we are seeing for some requests.


szabozoltan69 (approved):
Great change and should be really useful!

PR #1894, opened 2023-08-25
fetch molnix countries paginated to check if we were missing some cou…

batpad:
Maybe fixes the "invalid Country To" issue - try and ensure we fetch all countries from Molnix countries endpoint.

@szabozoltan69 would you be able to test this locally and we can deploy to staging?

cc @anamariaescobar


szabozoltan69:
Ok, I test it locally

szabozoltan69:
Deployed to Staging and Prod also

PR #1895, opened 2023-08-25
add setting so django respects X_FORWARDED_PROTO headers from proxy

batpad:
@szabozoltan69 this should fix the issue where links generated by the API were http instead of https


frozenhelium:
@anamariaescobar This should fix the issue of missing icons in the DREF export! It will be available soon in the Prod. Thanks @batpad @szabozoltan69 for the speedy work!

PR #1906, opened 2023-08-30
process all positions from Molnix, not just open ones

batpad:
cc @szabozoltan69 

This makes a change to fetch all historical positions from Molnix, not just open ones. This is because currently if there are any updates made to historical positions, we do not get those updates on GO.


szabozoltan69:
What a beautiful code...

PR #1919, opened 2023-09-28
add admin2 code into vector tiles

geohacker:
May be addresses https://github.com/IFRCGo/go-frontend/issues/2905 but let's not merge until confirmed that's the issue @szabozoltan69 @batpad @tnagorra


szabozoltan69:
@geohacker @batpad @tnagorra If you are ok with this (really harmless) change, let me merge and deploy it to Staging.

Issue #1927, opened 2023-10-13
Learnings Table and API Endpoints


ypyelab:
**Is your feature request related to a problem? Please describe.**
No. We want to have GO API endpoints for displaying operational learnings coming from DREF and Emergency Appeals documents. 

**Describe the solution you'd like**
We would like to have:
1. GO API GET, PUT, POST, PATCH, DELETE endpoints for the learnings table (table structure including column names and types attached). We will set up an external system to extract and categorize learnings from appeals documents and would like to post the results (categorized learnings) to be stored in GO. Therefore, we would like these API endpoints so that we can get/add/modify/delete this data. 
2. GO ADMIN learnings section. We would like to surface this table in the GO admin site, as we would like to have admin users manually validate the learnings and categorizations. As part of the validation, the users should be able to modify the learning and categories. These modified learnings/categories values should be stored in the table columns 'learning_validated', 'type_validated', 'sector_validated', and 'per_component_validated'.

**Additional context**
This is part of the ops learnings project, which tries to display operational learning data in different platforms/sections. The first step is to have the data easily retrievable (i.e., via GO API).

**Where does the data come from and how is it maintained?**
Externally generated from appeal documents and posted to GO via a POST endpoint.

**Should there be any restrictions around viewing or updating information related to this feature**
- GET endpoint: Each learning row is associated with an appeal (MDR code), the learning should have the same visibility as the appeal. 
- The other endpoints (PUT, POST, PATCH, DELETE) should be restricted.

**Links to documentation**
Table structure:
[go_ops_learning_endpoint.xlsx](https://github.com/IFRCGo/go-api/files/12892633/go_ops_learning_endpoint.xlsx)
Context of project:
https://docs.google.com/presentation/d/1HiulBBMZLoaazieX9lDuu13XRE6ZGG-RBkRoWBlafcY/edit?usp=drive_link

@tovari @AlexxxH

szabozoltan69:
**Permissions**
- not logged-in + logged-in without any group membership: can get the opslearning API endpoint that displays validated learnings
-  superusers + logged-in with OpsLearning Admin membership: can get/put/post/delete in the opslearning API endpoint that displays validated and not validated learnings)

szabozoltan69:
The fully functional opslearning API endpoint is deployed to Staging.
https://goadmin-stage.ifrc.org/api/v2/ops-learning/

ypyelab:
Context:
[...]

PR #1930, opened 2023-10-24
Ops Learning table

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1927


per/admin.py
```diff
@@ -155,6 +155,26 @@ class FormComponentQuestionAndAnswerAdmin(admin.ModelAdmin):
     pass
 
 
+class OpsLearningAdmin(admin.ModelAdmin):
+
+    def get_fields(self, request, obj=None):
```
batpad, line 160: The problem with this approach is that the user will have to select `is_validated` then Save, then reload the form to see the `_validated` fields - I think the desired user behaviour here would be to immediately show the fields when the user selects the check-box. That would require some hacky javascript to do - let's clarify the desired behaviour from @ypyelab exactly and decide what to do here.

per/models.py
```diff
@@ -656,3 +657,32 @@ class PerWorkPlan(models.Model):
...
+class LearningType(models.IntegerChoices):
+    LESSON_LEARNED = 1, _('Lesson learned')
+    CHALLENGE = 2, _('Challenge')
+
+
+@reversion.register()
+class OpsLearning(models.Model):
+    learning = models.TextField(verbose_name=_("learning"), max_length=900, null=True, blank=True)
+    learning_validated = models.TextField(verbose_name=_("learning (validated)"), max_length=900, null=True, blank=True)
+    appeal_code = models.CharField(verbose_name=_("appeal (MDR) code"), max_length=20, null=True, blank=True)
```
batpad, line 671: I wonder if we should maybe have another field called `appeal` that's an actual foreign key to Appeals? This is also fine, just thinking out loud ..

per/admin.py
```diff
@@ -155,6 +155,26 @@ class FormComponentQuestionAndAnswerAdmin(admin.ModelAdmin):
     pass
 
 
+class OpsLearningAdmin(admin.ModelAdmin):
+
```
batpad, line 159: Let's also add some other nice things for the Admin like `search_fields`, `list_filter` and `list_display` so we improve the experience around searching these in the Admin.

per/models.py
```diff
@@ -656,3 +657,32 @@ class PerWorkPlan(models.Model):
...
+
+
+class LearningType(models.IntegerChoices):
+    LESSON_LEARNED = 1, _('Lesson learned')
+    CHALLENGE = 2, _('Challenge')
+
+
+@reversion.register()
+class OpsLearning(models.Model):
+    learning = models.TextField(verbose_name=_("learning"), max_length=900, null=True, blank=True)
```
batpad, line 669: Is there a reason for the `max_length` here? a `max_length` for CharFields (which get created as `varchar` in the db) is required, but for `TextField` which gets created as a `text` field in postgres, I don't think there's any performance reasons to add a `max_length` - the only thing (that I'm aware of) this would really do is add validation in the Admin to not allow people to save anything over 900 chars.

If there's a business requirement for that, that's great, otherwise let's not add unneeded constraints.

per/admin.py
batpad: Based on chat with @ypyelab in Teams:

Let's show all the fields all the time and just make the non- `_validated` fields not editable in the admin. The way to do this would be to add the fields that we don't want editable in the Admin to a `readonly_fields` array in the Admin definition like here: https://stackoverflow.com/a/3967891 and I think that should be all we need.

szabozoltan69:
Is it possible to use a "conditionally readonly" field?

batpad:
> Is it possible to use a "conditionally readonly" field?

There probably is, but it might get a bit complex - what's the use case for the "conditionally" ? Right now as I understand it the non `_validated` fields should just always be readonly the admin? Setting it readonly in the admin will not affect the ability to write to it via the API.

szabozoltan69:
Thanks @batpad, now it's clearer. (I thought that the aim is to be able to push through the whole workflow on the Admin interface, that is why I had the impression that the non-validated fields should be writable at first and read-only after the is_validated is checked.)

batpad:
@szabozoltan69 that's a good consideration - if we do this, people will not be able to add new records in the Admin and there will be no way for users to either add new records or edit the non- `_validated` fields, except via the API.

@ypyelab do you think that's okay?

per/models.py
szabozoltan69: Done, 900 chars limit removed.

per/admin.py
szabozoltan69: Done, some example entities added – could be changed when the final structure is found out.

per/models.py
szabozoltan69: Good catch. Should be aligned with @ypyelab .

ypyelab:
Hi @szabozoltan69, @batpad, 

Ideally, it would be nice if people with the right access levels were able to add new records from the admin site. In such a case, of the creation of a new entry, could we only display the 'unvalidated' fields? And when they want to modify such an entry, we set up a similar behavior as we have defined before, showing them the 'unvalidated' and 'validated' fields, but allowing them only to modify the 'validated' fields.
Would that be possible?

szabozoltan69:
Let's continue at https://github.com/IFRCGo/go-api/pull/1950 please, to avoid this mass.
@batpad @ypyelab

ypyelab:
Thank you @szabozoltan69, we went through the admin site with the current 'validation' team, and they provided this feedback for the admin site:
- Have a 'Save and Next' button: For the persons that are validating, to not have to come back to the admin table of ops learning. Ideally, it goes to the next non-validated entry.
- Boxes of selected PER categories: In the list of PER components, not only put the highlighted components in the list but also show the selected categories, in boxes above the table. This request is done, in order to reduce errors, given the long-sized list.
- Preview in the view of all ops learning entries, of whether each entry has been validated or not (i.e. to see the validated variable at https://goadmin-stage.ifrc.org/en/admin/per/opslearning/). Something similar to the 'needs confirmation?' column at the appeal table in the admin site. And there was a related request (if possible), on not only 'viewing' whether it the entry has been validated, but also 'editing' if it has been validated from that view.
- Add the link to the DREF Final report. We have the MDR code already, but the team was wondering if we link it to the DREF Final Report ID in GO we would be able to 'click' directly to access the report.

szabozoltan69:
@ypyelab – thanks to the effective help of @thenav56, 3 of the 4 requests are ready.
Try it on https://goadmin-stage.ifrc.org/en/admin/per/opslearning/
The link to the DREF final report is still in progress and not ready, but the other 3 is there:
kép
kép
From https://goadmin-stage.ifrc.org/en/admin/per/opslearning/1/change/ "Save and next one" leads to https://goadmin-stage.ifrc.org/en/admin/per/opslearning/2/change/
Also the nicer multiple select is done via autocomplete (!). So you can just type and get what you want. (Or choose, if you are lazy to type.)

PR #1931, opened 2023-10-25
Different resource allocation for production and staging

sunu:
Uses different helm values files based on environment setting to set allocated resources

cc @batpad

### DIFF
--- a/deploy/helm/ifrcgo-helm/values-production.yaml
+++ b/deploy/helm/ifrcgo-helm/values-production.yaml
@@ -1,0 +1,35 @@
+api:
+  resources:
+    requests:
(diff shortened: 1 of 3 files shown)


batpad:
@sunu this is so good!

Am hoping we can use the same approach for other values that need to be different for staging and prod, but are not secrets. But we can do that as a separate PR.

Let us know when you feel this is ready to test, and I think we can test this on staging - I would really like to have this in place when we do the production deploy so we can be more flexible about resource allocation between prod and staging.

cc @szabozoltan69

szabozoltan69:
Great progress!!!

PR #1948, opened 2023-11-13
Create TLS secrets from own certs

sunu:
## Changes
- the contents of the crt and key files should be stored as base64 encoded strings in the respective azure pipeline secrets
- removes cert-manager annotation from ingress because we are not using letsencrypt certs anymore

cc @batpad @szabozoltan69


batpad:
@sunu nice ! @szabozoltan69 when you're around, let's test this on staging ..

Currently, we use the same wildcard subdomain cert for all the domains (staging, prod and the prod add-on domain), but I think it's good to keep them separately in the secrets in case that changes down the line.

szabozoltan69:
Ok, I add the needed pipeline secrets.

szabozoltan69:
@sunu @batpad The pipeline secret vars are added.

sunu:
@szabozoltan69 awesome! Marking the PR ready so that we can merge and test. If anything breaks, happy to jump on a call to debug and fix. Also, just to make sure, the contents of the secrets should be base64 encoded.

szabozoltan69:
Yes, everything is base64 encoded. Thanks @sunu really!
After a short staging test I'll merge it to master (prod) also, to avoid crash of letsencript overtime there (and due to the recent habit that we keep parallel develop and master as long as quick fixes are needed).

szabozoltan69:
@sunu I see no change in deploy/helm/ifrcgo-helm/templates/config/*, I hope it is OK like this.

sunu:
@szabozoltan69 Yea, the certs need separate `tls` type secrets of their own. That's why I didn't add them to the existing secret in deploy/helm/ifrcgo-helm/templates/config/secret.yaml

szabozoltan69:
(The first deployment died due to conflicting (already existing) resource, but I gave a manual
kubectl delete secret ifrcgo-helm-secret-api
command, then it's ok now.)

PR #1958, opened 2023-11-17
Go to next object (after Save) in Django Admin

szabozoltan69:
Making OpsLearning nicer.


per/admin.py
```diff
@@ -155,10 +156,36 @@ class FormComponentQuestionAndAnswerAdmin(admin.ModelAdmin):
...
+            if qs:
+                response['Location'] += '?' + qs
+            return response
+
+        return super().response_change(request, obj)
+# stackoverflow.com/questions/58014139/how-to-go-to-next-object-in-django-admin
+
+    def render_change_form(self, request, context, add=False, change=False, form_url='', obj=None):
+        context.update({'show_save_and_next': True})
+        return super().render_change_form(request, context, add, change, form_url, obj)
```
thenav56, line 182: We can move this to the Mixin as well.

For 'admin:per_opslearning_' we can use
```
        app_label = obj._meta.app_label
        model_name = obj._meta.model_name
        f'admin:{app_label}_{model_name}_change'
        f'admin:{app_label}_{model_name}_changelist'
```

thenav56 (changes requested):

per/admin.py
szabozoltan69: Hi @thenav56 , I couldn't make a working version by moving response_change to the mixin.

(Even not by leaving this in OpsLearningAdmin:  )
```
    def render_change_form(self, request, context, add=False, change=False, form_url='', obj=None):
        context.update({'show_save_and_next': True})
        return super().render_change_form(request, context, add, change, form_url, obj)
```

PR #1970, opened 2023-12-02
OpsLearning API endpoint

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1927

### DIFF
--- a/per/drf_views.py
+++ b/per/drf_views.py
@@ -57,8 +57,9 @@ from .serializers import (
     PublicPerProcessSerializer,
     PublicPerAssessmentSerializer,
     OpsLearningSerializer,
+    PublicOpsLearningSerializer,
 )
-from per.permissions import PerPermission
+from per.permissions import PerPermission, OpsLearningPermission
 from per.filter_set import (
     PerOverviewFilter,
     PerPrioritizationFilter,
@@ -372,4 +373,11 @@ class OpsLearningViewset(viewsets.ModelViewSet):
     """
     queryset = OpsLearning.objects.all()
     serializer_class = OpsLearningSerializer
-    permission_classes = [permissions.IsAuthenticated]
+    permission_classes = [OpsLearningPermission]
+
+    def get_serializer_class(self):
+        user = self.request.user
+        if user.is_superuser or user.groups.filter(name="OpsLearning Admin").exists():
+            return OpsLearningSerializer
+        else:
+            return PublicOpsLearningSerializer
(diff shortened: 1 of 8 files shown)


thenav56 (changes requested):
Looks good. Few comments

per/drf_views.py
```diff
@@ -362,3 +364,12 @@ def get_queryset(self):
                 "-assessment_number").distinct("country_id").values("id")
         )
         return self.get_filtered_queryset(self.request, queryset, dispatch=0)
+
+
+class OpsLearningViewset(viewsets.ModelViewSet):
+    """
+    A simple ViewSet for viewing and editing OpsLearning records.
+    """
+    queryset = OpsLearning.objects.all()
```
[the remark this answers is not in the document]
thenav56: We also have PublicOpsLearningSearializer, so that means normal/guest can also see them, right?

We probably need to update our permission module for each operation GET/POST/PATCH/DELETE where GET is for everyone with different serializer and POST/PATCH/DELETE is for admin/ops-maintainer.
```python
if request.method in permissions.SAFE_METHODS:
    # Check permissions for read-only request
else:
    # This will be POST/PATCH/DELETE
    # Check permissions for write request
```

To specify which fields are read only, we can use this https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields

per/drf_views.py
```diff
@@ -362,3 +365,19 @@ def get_queryset(self):
...
+    serializer_class = OpsLearningSerializer
+    permission_classes = [IsAuthenticated, OpsLearningPermission]
+
+    def get_serializer_class(self):
+        user = self.request.user
+        if not user.is_superuser:
+            return OpsLearningSerializer
+        else:
+            return PublicOpsLearningSerializer
```
thenav56, line 383: I am a bit confused here, is this logic for superuser or just logged-in? Maybe something like this?
```suggestion
        if not user.is_authenticated:
            return PublicOpsLearningSerializer
        return  OpsLearningSerializer
```

thenav56 (changes requested):
Added feedback on the comment

per/drf_views.py
szabozoltan69: You were correct, it was erroneous: 'if **not** user.is_superuser'.
I fixed the mistakes and tried to implement https://github.com/IFRCGo/go-api/issues/1927#issuecomment-1840510193 permission requirements.

szabozoltan69:
Thanks so much for the refactors @thenav56 !

PR #1971, opened 2023-12-05
Save disk config in a k8s configmap

sunu:
## Changes

Saves disk config in a k8s configmap.

We will use this configmap as the intermediate link between terraform and helm where we store terraform artifacts used in the go-api helm chart. Later on we will use this configmap in helm when we decouple the go-api helm chart from terraform


szabozoltan69:
Should we merge it @batpad?

batpad:
> Should we merge it @batpad?

Yes. Sorry for not getting to this earlier -- we can also merge tomorrow morning if that's better.

Issue #1885, opened 2023-08-24
Backend Infrastructure next steps (EPIC)


batpad:
We successfully migrated the GO deploy to a Kubernetes based backend on AKS with CI running using Azure Pipelines: https://github.com/IFRCGo/go-api/issues/1531

Creating this ticket with further actions and next steps here. Some of these are smaller, more immediate tasks, and some of these are more long-term with a view to expanding the infrastructure setup to include other GO-adjacent applications like the Risk Module, as well as other projects like Alert Hub and other experimental and student projects.

More short-term tasks:

 - [ ] Setup SSH tunnel to database to allow database access for external users
 - [ ] Setup resource configuration such that we can have different resource allocations for staging and production
 - [ ] Improve documentation for new deploy setup, add guidelines on creating releases, etc. under new system
 - [ ] Setup Sentry within the cluster

Medium-term:
 - [ ] Move `deploy` folder in go-api to its own repository, separating out management of the cluster and deployment of individual applications
 - [ ] Setup CPU / Memory utilization based auto-scaling on the GO API
 - [ ] Improve handling of secrets and configuration variables to make it easier for individual applications to manage secrets and configuration
 - [ ] Create consistent pattern to deploy go-api, risk module, and other applications
 - [ ] Add Schema JSON for values configuration to make validation of configuration options better
 - [ ] Add basic tests for Helm chart to increase confidence while deploying applications

Longer-term:
 - [ ] Setup a `playground` cluster to allow more experimental projects to use as a testing space for deployments and ideas
 - [ ] Have comprehensive documentation on best practices for deploying an app to the "GO Ecosystem", including setting up CI / CD
 - [ ] Investigate and setup comprehensive monitoring and logging tools across clusters, manage permissions so that we can give well scoped permissions to people to monitor their own applications

@tovari I can break these out into separate more detailed tickets as we figure prioritization.

cc @szabozoltan69 @geohacker - please add anything here that I might have missed!

+cc @sunu

batpad:
+cc @k9845 - there are some dependencies here that ideally we'd get through before doing the Risk Module deploy into the cluster, so we can do it in a more "proper way", but again, let's discuss how best to proceed - there maybe an intermediary solution if we want to hasten up the migration of Risk Module onto the GO cluster.

szabozoltan69:
[...]

PR #1974, opened 2023-12-07
Define helm values directly in the values files

sunu:
@batpad I have defined all the values that gets passed to the configmap for staging. Can you do the same for production?

~I have added some of the keys from the `ifrcgo-helm-api-secret` secret which I think don't have to be secrets but have left the values undefined. @batpad could you double check that all of them are safe to be stored in the repo as text?~ (we decided to leave the secret values in one place and not move them over to helm values)

Do we have an issue for tracking the work related to deployment code decoupling and migration that we can tag in these PRs?


deploy/helm/ifrcgo-helm/values-production.yaml
```diff
@@ -1,4 +1,26 @@
...
+  ELASTIC_SEARCH_HOST: elasticsearch://ifrcgo-helm-elasticsearch:9200
+  FRONTEND_URL: go.ifrc.org
+  GO_ENVIRONMENT: production
+  IFRC_TRANSLATION_DOMAIN: https://microservices-staging.ifrc.org
+  SENTRY_DSN: https://342632f67ea942c9bde214c0bb7e91be@sentry.northeurope.cloudapp.azure.com/2
+  SENTRY_SAMPLE_RATE: "0.2"
+
 api:
+  domain:
+  additionalDomain:
```
sunu, line 23: @batpad could you fill in these values?

batpad: @sunu filled in `domain` and `additionalDomain` - let me know if you need me to fill in any others.

batpad:
> Do we have an issue for tracking the work related to deployment code decoupling and migration that we can tag in these PRs?

@sunu if you feel it helps to make a more specific ticket, let's work on that together - for now I think you can link to the master infra ticket: https://github.com/IFRCGo/go-api/issues/1885

Issue #1975, opened 2023-12-08
Add new surge alert filters to API


JonathanGarro:
We added a bunch of surge alert fields earlier this year from RRMS. I'm hoping we can make those accessible via the API for end users like myself. Per the [documentation](https://goadmin.ifrc.org/docs/#tag/api/operation/api_v2_surge_alert_list), it looks like there are several missing from what we added (as well as some that I believe were already there), including: 

- `start` (as well as `start__gt`, `__gte`, `__lt`, `__lte`)
- `molnix_status`
- `message`
- `country`
- `molnix_tags` - these have the new "groups" key that contain a bunch of keys i'd love to have access to, such as `ROLES`, `LANGUAGE`, `SECTOR`, `rotation`, etc.

szabozoltan69:
Ok, so we should append the needed fields to the filterset, to this:
https://github.com/IFRCGo/go-api/blob/develop/notifications/drf_views.py#L23_L26

szabozoltan69:
Is it ok if I add these?
molnix_id, molnix_status, start, message, country

JonathanGarro:
is it possible to have it filter those molnix_tags too? or is the structure too complex?

batpad:
FYI: Renamed the title of this ticket to "Add new surge alert **filters** to API" instead of "Add new surge alert fields to API" to make the requirement a bit clearer.

szabozoltan69:
Also in docs:
![kép](https://github.com/IFRCGo/go-api/assets/6050643/8d82c30d-9a49-4b0d-b916-648233daec2f)

szabozoltan69:
Deployed to Staging.
Many thanks to @thenav56 for the great cleanup!

szabozoltan69:
https://goadmin-stage.ifrc.org/api/v2/surge_alert/?country__iso3__in=HND,HUN&molnix_tag_names=LOGS,COM,LOG-CO

JonathanGarro:
thanks @tovari - i just tested a few and it seems to work great! huge thanks to the team, this will be extremely useful! i appreciate y'all.

PR #1976, opened 2023-12-08
New filters on SurgeAlerts

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1975

### DIFF
--- a/notifications/drf_views.py
+++ b/notifications/drf_views.py
@@ -1,7 +1,8 @@
-from datetime import datetime, timedelta, timezone
-from django.db.models import Q
+# from datetime import datetime, timedelta, timezone
+# from django.db.models import Q
 from django_filters import rest_framework as filters
 from django_filters.widgets import CSVWidget
+from main.filters import CharInFilter
 from rest_framework.authentication import TokenAuthentication
 from rest_framework.permissions import IsAuthenticated
 from rest_framework import viewsets
@@ -9,13 +10,9 @@ from .models import SurgeAlert, Subscription
 from deployments.models import MolnixTag
 from .serializers import (
     SurgeAlertSerializer,
-#   UnauthenticatedSurgeAlertSerializer,
+    # UnauthenticatedSurgeAlertSerializer,
     SubscriptionSerializer,
 )
-
-
-class CharInFilter(filters.BaseInFilter, filters.CharFilter):
-    pass
 
 
 class SurgeAlertFilter(filters.FilterSet):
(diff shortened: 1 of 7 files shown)


notifications/drf_views.py
```diff
@@ -1,29 +1,67 @@
...
+
+    def molnix_tags_filter(self, qs, name, value):
+        if value:
+            return qs.filter(molnix_tags__in=value).order_by('id').distinct()
+        return qs
+
+    def molnix_tag_names_filter(self, qs, name, value):
+        if value:
+            return qs.filter(molnix_tags__name__in=value).order_by('id').distinct()
+        return qs
```
thenav56, line 48: I checked and looks like we can also do this in this way.

For the order_by('id'), we shouldn't do ordering in filters as we have separate module for ordering.

We should move `CharInFilter` to a `api/utils.py` or maybe create a new one `main/filters.py`
```suggestion
    molnix_tags = filters.ModelMultipleChoiceFilter(
        label='tag-ids',
        field_name='molnix_tags',
        lookup_expr='in',
        help_text='Molnix_tag GO identifiers, comma separated',
        widget=CSVWidget,
        queryset=MolnixTag.objects.all(),
    )
    molnix_tag_names = CharInFilter(
        label='tag-names',
        field_name='molnix_tags__name',
        lookup_expr='in',
        help_text='Molnix_tag names, comma separated',
        widget=CSVWidget,
    )

    @property
    def qs(self):
        return super().qs.distinct()
```

notifications/tests.py
```diff
@@ -56,3 +58,75 @@ def test_surge_translation(self):
...
+            description='Americas Region',
+            tag_type='regular',
+            tag_category='molnix_region')
+        surge_alert_1 = SurgeAlert.objects.create(
+            message='CEA Coordinator, Floods, Atlantis', atype=6, category=2, molnix_status='active', country=country_1)
+        surge_alert_1.molnix_tags.set([molnix_tag_1, molnix_tag_2])
+        surge_alert_2 = SurgeAlert.objects.create(
+            message='WASH Coordinator, Earthquake, Neptunus', atype=5, category=1, molnix_status='active', country=country_2)
+        surge_alert_1.molnix_tags.set([molnix_tag_1, molnix_tag_2])
+        surge_alert_2.molnix_tags.set([molnix_tag_1, molnix_tag_3])
```
thenav56, line 93: We can also use factoryboy here.

notifications/tests.py
```diff
@@ -56,3 +58,75 @@ def test_surge_translation(self):
...
+             response['results'][1]['molnix_tags'][0]['name'],
+             response['results'][1]['molnix_tags'][1]['name']},
+            {'OP-6700', 'L-FRA', 'AMER'})
+
+        # we have 2 SurgeAlerts with MolnixTags A, B and A, C.
+        # Filtering for A, B, C gives both. Filtering for only B or C gives 1.
+
+        # filtering by molnix_tags
+        response = self.client.get('/api/v2/surge_alert/?country__iso3__in=ATL,NPT&molnix_tags=' + str(molnix_tag_2.id)
+            ).json()
```
thenav56, line 116: We can pass GET arguments as dict.
https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.get

```suggestion
        def _to_csv(*items):
            return ','.join([str(i) for i in items])

        response = self.client.get(
            '/api/v2/surge_alert/'
            {
                'country__iso3__in': 'ATL,NPT',
                'molnix_tags': _to_csv(molnix_tag_2.id),
            },
        ).json()
```

thenav56 (changes requested):
Looks good. Few suggestions

notifications/tests.py
szabozoltan69: I'd keep it this way, due to there is not yet factory in this notifications app and it would be a huge work.
Not mentioning the experience that factoryboy sometimes forces snapshot-test regenerating without any real reason.

thenav56: I meant just the factories, without snapshot-test. I will push this changes.

szabozoltan69: I am not against factories (mainly if it's not me who makes it), but sometimes there is some relevance the "ambiance of real strings" in tests, I mean OP-6700 as molnix-tag instead of jUierwKLJ. It can be easier to understand the purpose of the tests with human-readable strings.

thenav56 (approved):

Issue #2007, opened 2024-01-17
[PROD] Ingest Appeal Docs


szabozoltan69:
## Issue
Recently the ingest_appeal_docs job does not run fine.
Without a header hack (of personal cookie data) the scraping of www.ifrc.org/appeals/ gives:
'reason': 'Forbidden',
'status': 403,

The bigger issue is that why is that failure invisible in cronjob items? Normally an erroneous run should be seen there and give a big warning message.

@batpad @thenav56

szabozoltan69:
We could use such API endpoints instead of scraping:
https://go-api.ifrc.org/api/PublicSiteAppeals?Appealnumber=MDRKZ012&Hidden=false

tovari:
@arunissun, would you mind to compare staging and prod data of the 'appeal_document' endpoint?

szabozoltan69:
Some differences can be; prod and staging count:
```
postgres=> select count(*) from api_appealdocument;
 count 
-------
  8126

postgres=> select count(*) from api_appealdocument;
 count 
-------
  8133
```

arunissun:
@tovari I will check the staging and prod data for appeal documents

tovari:
Thanks @arunissun for the analysis. Looks like the scraper doesn't work well anymore on prod (there are missing docs from 2024).


![Image](https://github.com/IFRCGo/go-api/assets/68283148/8de1dd5d-06b4-40eb-a049-273dc1f7fc87)

PR #2014, opened 2024-01-22
Add no-scraping appealdocument reader

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2007
If this job is ok, no need for the two other ingest_* jobs for appealdocuments.


szabozoltan69:
@batpad - the two old appealdoc ingester could be left, and this new one is ok for both (without scraping anything).

Issue #1990, opened 2024-01-09
[PROD] Django upgrade to 4.2


szabozoltan69:
## Issue
According to https://endoflife.date/django the end of support of the recent 3.2 Django LTS version is in April 2024.
So (if possible, in March) we should upgrade to Django 4.2 LTS.
@batpad @thenav56 @k9845

szabozoltan69:
```
-python = "^3.8"
+python = "^3.9"

-Django = ">=3.2,<3.3"
+Django = ">=4.2,<5.0"

-django-reversion-compare = "==0.9.0"
+django-reversion-compare = "==0.16.2"

-django-reversion = "==3.0.5"
+django-reversion = "==5.0.12"

-django-tinymce4-lite = "==1.7.4"
+django-tinymce = "==3.7.1"

-djangorestframework = "==3.11.2"
+djangorestframework = "==3.14.0"
```

szabozoltan69:
```
pip list|grep jango
Django                            4.2.10
django-admin-autocomplete-filter  0.7.1
django-admin-list-filter-dropdown 1.0.3
django-cors-headers               3.11.0
django-coverage                   1.2.4
django-debug-toolbar              4.1.0
django-enumfield                  2.0.2
django-environ                    0.8.1
django-extensions                 2.0.6
django-filter                     2.4.0
django-guardian                   2.4.0
django-haystack                   3.2.1
django-modeltranslation           0.17.5
django-read-only                  1.12.0
django-redis                      5.0.0
django-reversion                  5.0.12
django-reversion-compare          0.16.2
django-storages                   1.11.1
django-stubs                      4.2.7
django-stubs-ext                  4.2.7
django-tinymce                    3.7.1
djangorestframework               3.14.0
djangorestframework-camel-case    1.2.0
djangorestframework-csv           2.1.1
djangorestframework-guardian      0.1.1
graphene-django                   2.16.0
opencensus-ext-django             0.7.4
pytest-django                     4.8.0
```

PR #2031, opened 2024-02-06
Django upgrade to 4.2

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1990


szabozoltan69:
Still breaks with:
django.template.library.InvalidTemplateLibrary: Invalid template library specified. ImportError raised when trying to load 'rest_framework.templatetags.rest_framework': cannot import name 'parse_header' from 'django.http.multipartparser' (/usr/local/lib/python3.9/site-packages/django/http/multipartparser.py)

docker-compose.yml
```diff
@@ -61,7 +61,7 @@ x-server: &base_server_setup
 
 services:
   db:
-    image: postgis/postgis:11-3.1-alpine
+    image: postgis/postgis:12-3.4-alpine
```
thenav56, line 64: Hey @szabozoltan69, doing this might break current setup. Mabye we should use different directory instead for the data? or use a volume instead? 
We will also need to specify how to move data from old db version to new one. 
```
./.db-12/pg:/var/lib/postgresql/data
```

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8.12-bullseye
+FROM python:3.9-bullseye
```
thenav56, line 1: Just wondering if you have tried newest version?  (Latest stable one is 3.12)

3.9 is also nice, it supports https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

api/serializers.py
```diff
@@ -2061,7 +2061,7 @@ class SearchEmergencySerializer(serializers.Serializer):
 class SearchSurgeAlertSerializer(serializers.Serializer):
     id = serializers.IntegerField()
     name = serializers.CharField()
-    keywords = serializers.ListField(serializers.CharField())
+    keywords = serializers.ListField()
```
thenav56, line 2064: Hey @szabozoltan69, Any reason why we are removing child definition here?
Maybe we need to use child=serializers.CharField() instead?

thenav56 (changes requested):
Thanks for the upgrade :D
Few comments/questions

szabozoltan69:
Thanks for the great catches, @thenav56 , all of them are relevant and useful, I'll fix them soon!

docker-compose.yml
szabozoltan69: Great idea! (Maybe I'll use .db/pg-12)
The new db can be done via a dump and a load, like this:
```
pg_dump -O    -c --if-exists -h 127.0.0.1 -U test test > /tmp/testdump.sql
```
Then the load is via a simple \i ...testdump.sql from the new db.

api/serializers.py
szabozoltan69: Placed back (and the other places also) as
```
...  = serializers.ListField(child=serializers.CharField())
```

Dockerfile
thenav56: Thanks for checking

docker-compose.yml
thenav56: Nice... let's add this to PR description as well

thenav56 (approved):

szabozoltan69:
Also python-magic = "==0.4.27" to packages.

szabozoltan69:
Hi  @k9845 , I think we should let this branch go and initiate another branch for Django 4.2 cutover. It would be hard to keep this one fine and usable (and not overwriting anyting). @thenav56 – what do you think?

szabozoltan69:
Will be closed.
Reimplemented via https://github.com/IFRCGo/go-api/pull/2152

PR #2034, opened 2024-02-08
Migrate legacy data from django reversion

thenav56:
Addresses
- https://github.com/IFRCGo/go-web-app/issues/540

## Changes

- Migrate Per Overview version datetime data to date to avoid Django serialization error
- Add helper function for future


szabozoltan69:
@thenav56 Thanks for this.
Maybe sometimes it is enough simply to delete previous reversion data and create new initial ones in such cases, like:
```
./manage.py deleterevisions per.opslearning
./manage.py createinitialrevisions --comment="Reinitializing" per.opslearning
```

thenav56:
Hey @szabozoltan69, This should fix by updating existing data by creating a new migration something like this
https://github.com/IFRCGo/go-api/blob/c261022ec67138a28ee45ed310ca38a50f74f890/per/migrations/0098_fix_reversion_data_20240208_0502.py#L1-L34

If it is okay from user's perspective to remove them, then the above-mentioned command will be a quicker way to fix them.

szabozoltan69 (approved):

PR #2058, opened 2024-02-27
Tagging OpsLearning via a cron job

szabozoltan69:
Addresses https://github.com/IFRCGo/go-api/issues/1927


szabozoltan69:
In the Django 4.2 upgrade these 2 new packages are already there (nltk, rertrying). So the conflicts in poetry.lock (and pyproject.toml) can be ignored and simply omitted from the PR, if we merge it after the Django 4.2 cutover.

PR #2068, opened 2024-03-12
Fix/country page

k9845:
Addresses xxxxxx

## Changes

* ...

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


api/models.py
```diff
@@ -2492,6 +2492,11 @@ class ExportType(models.TextChoices):
         null=True, blank=True,
         on_delete=models.SET_NULL,
     )
+    per_country = models.IntegerField(
+        verbose_name=_('Per Country Id'),
+        null=True,
+        blank=True
+    )
```
thenav56, line 2499: Why not use ForeignKey?

api/serializers.py
```diff
@@ -2378,7 +2378,10 @@ def create(self, validated_data):
         else:
             title = "Export"
         user = self.context['request'].user
-        validated_data['url'] = f'https://{settings.FRONTEND_URL}/{export_type}/{export_id}/export/'
+        if export_type == Export.ExportType.PER:
+            validated_data['url'] = f'https://{settings.FRONTEND_URL}/countries/{country_id}/{export_type}/{export_id}/export/'
+        else:
+            validated_data['url'] = f'https://{settings.FRONTEND_URL}/{export_type}/{export_id}/export/'
```
thenav56, line 2384: Let's handle all the cases here. For the last else - we should throw an error instead of doing a default else.

api/serializers.py
```diff
@@ -2378,7 +2378,10 @@ def create(self, validated_data):
         else:
             title = "Export"
         user = self.context['request'].user
-        validated_data['url'] = f'https://{settings.FRONTEND_URL}/{export_type}/{export_id}/export/'
+        if export_type == Export.ExportType.PER:
+            validated_data['url'] = f'https://{settings.FRONTEND_URL}/countries/{country_id}/{export_type}/{export_id}/export/'
```
thenav56, line 2382: We should generate this URL on the export task instead in the serializer.

databank/management/commands/ingest_climate.py
```diff
@@ -13,66 +13,42 @@
...
-                                precipitation=precipitation_monthly
+                                **data
                             )
-                        logger.info('Minimum, maximum and Average temperature of country temperature data added successfully')
-
-                except Exception as ex:
-                    logger.error(f'Error in ingesting climate data: {ex}')
-                    continue
+            except Exception as ex:
+                logger.error(f'Error in ingesting climate data: {ex}')
```
thenav56, line 53: *
```suggestion
                logger.error(f'Error in ingesting climate data:', exc_info=True)
```

databank/management/commands/ingest_worldbank.py
```diff
@@ -78,6 +78,7 @@ def handle(self, *args, **kwargs):
                                     if existing_data is None or existing_data[1] < year:
                                         country_dict[geo_id].append((pop, year, indicator))
                                         logger.info(json.dumps(country_dict))
+                                        print(country_dict)
```
thenav56, line 81: *
```suggestion
```

deploy/helm/ifrcgo-helm/values.yaml
```diff
@@ -119,8 +119,7 @@ cronjobs:
     schedule: '1 0 * * *'
   - command: 'update_surge_alert_status'
     schedule: '1 */12 * * *'
-  - command: 'ingest_gdacs'
-    schedule: '1 */12 * * *'
```
thenav56, line 123: Do we still need api/management/commands/ingest_gdacs.py ?

per/serializers.py
```diff
@@ -326,7 +342,28 @@ def get_assessment_ratings(self, overview):
...
+        per_admin_region_id = [
+            codename.replace('per_region_admin_', '')
+            for codename in Permission.objects.filter(
+                group__user=user,
+                codename__startswith='per_region_admin_',
+            ).values_list('codename', flat=True)
+        ]
+        per_admin_region_id = list(map(int, per_admin_region_id))
+
+        if country_id  in per_admin_country_id or region_id in per_admin_region_id or user.is_superuser or user.has_perm("api.per_core_admin"):
```
thenav56, line 366: Maybe we can define a function instead of doing this in two places.

susilnem (approved):

Issue #2077, opened 2024-03-20
[PROD] Image upload difficulty on Admin pages


szabozoltan69:
## Issue
On the Admin page, at countries or regions there are snippets.
Near the snippets there is an option for image upload, but if we try to use it, gives an [Error 500](https://porgeto.hu/tmp/20240320e.txt).
(There are similar snippets and image upload option for events (emergencies) also, but it works fine.)
In Sentry you can also find this item, id=1648.
Somehow the revisions, the translations, the encoding can be related.

szabozoltan69:
Somehow the country snippet object has no name:
```
(Pdb) p content_type
<ContentType: api | country snippet>
(Pdb) pp vars(obj)
{'_mt_disable': False,
 '_state': <django.db.models.base.ModelState object at 0x7f673a02c670>,
 'country_id': 16,
 'id': 102,
 'image': 'countries/2024/03/21/BEE_logo_PERNpnp.jpg',
 'position': 3,
 'snippet': None,
 'snippet_ar': None,
 'snippet_en': None,
 'snippet_es': None,
 'snippet_fr': None,
 'translation_module_original_language': 'en',
 'translation_module_skip_auto_translation': False,
 'visibility': 3}
(Pdb) p obj
(nothing)
```

szabozoltan69:
Solution found.
In case of "no HTML snippet part" (only image) no name can be added by Django.
So instead of
    def __str__(self):
        return self.snippet
We should use:
    def __str__(self):
        return self.snippet or self.image.url

PR #2078, opened 2024-03-20
Feature/check svg file type

szabozoltan69:
Validating file uploads.
Also fixes https://github.com/IFRCGo/go-api/issues/2077 (which issue was a blocker of proper testing).


per/serializers.py
```diff
@@ -541,6 +555,8 @@ class Meta:
         read_only_fields = ("created_by",)
 
     def create(self, validated_data):
+        if "file" in validated_data:
+            check(validated_data["file"])
```
thenav56, line 559: Same as other, We will need functional validation

per/serializers.py
```diff
@@ -995,13 +1011,14 @@ def validate(self, data):
             )
         return data
 
-
     def create(self, validated_data):
+        if "file" in validated_data:
+            check(validated_data["file"])
```
thenav56, line 1016: For create, We can ignore "file" in validated_data check as the `file is a required field in PerDocumentUpload.

per/serializers.py
```diff
@@ -222,6 +223,19 @@ class Meta:
...
+class NiceDocumentSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = NiceDocument
+        fields = "__all__"
+
+    def create(self, validated_data):
+        if "document" in validated_data:
+            check(validated_data["document"])
+        # validated_data["created_by"] = self.context["request"].user  # FIXME: this field would be fine.
```
thenav56, line 235:
```suggestion
```

flash_update/serializers.py
```diff
@@ -59,6 +58,8 @@ class Meta:
         read_only_fields = ('created_by',)
 
     def create(self, validated_data):
+        if 'file' in validated_data:
+            check(validated_data['file'])
```
thenav56, line 62: For create, We can ignore "file" in validated_data check as the `file is a required field in FlashGraphicMap.

dref/serializers.py
```diff
@@ -66,6 +67,8 @@ class Meta:
         read_only_fields = ("created_by",)
 
     def create(self, validated_data):
+        if "file" in validated_data:
+            check(validated_data["file"])
```
thenav56, line 71: For create, We can ignore "file" in validated_data check as the `file is a required field in DrefFile.

To make sure it is validated for update as well, a function validate_ will be easier.

```python
def validate_file(self, file):
    check(file)
    return file
```

dref/serializers.py
```diff
@@ -471,6 +474,8 @@ def validate_operation_timeframe(self, operation_timeframe):
         return operation_timeframe
 
     def create(self, validated_data):
+        if "budget_file_preview" in validated_data:
+            check(validated_data["budget_file_preview"])
```
thenav56, line 478: Same here, we will need functional validate instead

dref/serializers.py
```diff
@@ -617,6 +622,8 @@ def get_total_timeframe(self, start_date, end_date):
         return None
 
     def create(self, validated_data):
+        if "budget_file_preview" in validated_data:
```
thenav56, line 625: Same here, we will need functional validate instead

dref/serializers.py
```diff
@@ -927,6 +934,9 @@ def create(self, validated_data):
         # here check if there is operational update for corresponding dref
         # if yes copy from the latest operational update
         # else copy from dref
+
+        if "financial_report_preview" in validated_data:
```
thenav56, line 938: Same here, we will need functional validate instead

databank/serializers.py
```diff
@@ -83,6 +84,11 @@ class Meta:
         model = KeyDocument
         exclude = ('overview',)
 
+    def create(self, validated_data):
+        if 'file' in validated_data:
+            check(validated_data['file'])
```
thenav56, line 89: Same as other, We will need functional validation

api/models.py
```diff
@@ -647,7 +647,7 @@ class Meta:
         verbose_name_plural = _('region snippets')
 
     def __str__(self):
-        return self.snippet
+        return (self.snippet[:80] if self.snippet else None) or (self.image.url if getattr(self, 'image') else '')
```
thenav56, line 650: Without default value in `getattr`, it throws AttributeError error
```suggestion
        return (
            self.snippet[:80] if self.snippet else None
        ) or (
            self.image.url
            if getattr(self, 'image', None) and self.image.url
            else ''
        )
```

api/models.py
```diff
@@ -724,7 +724,7 @@ class Meta:
         verbose_name_plural = _('country snippets')
 
     def __str__(self):
-        return self.snippet
+        return (self.snippet[:80] if self.snippet else None) or (self.image.url if getattr(self, 'image') else '')
```
thenav56, line 727: Same as before

api/models.py
```diff
@@ -1020,7 +1020,7 @@ class Meta:
         verbose_name_plural = _('snippets')
 
     def __str__(self):
-        return self.snippet if self.snippet else ''
+        return (self.snippet[:80] if self.snippet else None) or (self.image.url if getattr(self, 'image') else '')
```
thenav56, line 1023: Same as before

utils/file_check.py
```diff
@@ -0,0 +1,19 @@
...
+    Raises:
+        ValidationError: If the file type is disallowed.
+    """
+
+    if file:
+        mime = Magic(mime=True)
+        file_type = mime.from_buffer(file.read(1024))
+        file.seek(0)  # Read a portion for analysis, then reset file pointer
+        if file_type == "image/svg+xml":
+            raise ValidationError("SVG image upload is not allowed. Please use another image type.")
```
thenav56, line 19: We will need to use a different name for this function.

Maybe, `validate_allowed_file_type`

Also, let's add a option for bool response
```suggestion
def validate_allowed_file_type(file, raise_exception=True) -> bool:
    """
    This function validates a file object based on a disallowed mimetype.
    Arg:
        file: The file object to validate.
    Raises:
        ValidationError: If the file type is disallowed.
    """

    if file:
        mime = Magic(mime=True)
        file_type = mime.from_buffer(file.read(1024))
        file.seek(0)  # Read a portion for analysis, then reset file pointer
        if file_type == "image/svg+xml":
            if raise_exception:
                raise ValidationError("SVG image upload is not allowed. Please use another image type.")
            return False
        return True
```

thenav56 (changes requested):
@szabozoltan69 Look good.

I noticed the validation using `file_field` in validated_data. It is much easier to have a function `validate_<file_field>` instead, as it will handles both create and update, and we don't have to check if the field is provided by the client for different API methods.

szabozoltan69:
@thenav56 I implemented the suggested changes. Really nicer this way! :-)

szabozoltan69:
(Also tested it locally.)

thenav56 (approved):

PR #2086, opened 2024-03-29
Build and push docker images and helm charts on every push to develop

sunu:
The CI will build appropriately tagged docker images and helm charts for every push to develop and for every tag. We will used these published images and charts to deploy go-api once we move deployment to a separate repo.

The docker images and Helm charts are published to GHCR.

refs https://github.com/IFRCGo/go-api/issues/1885

cc @batpad @szabozoltan69

### DIFF
--- a/.github/workflows/build-publish-docker-helm.yaml
+++ b/.github/workflows/build-publish-docker-helm.yaml
@@ -9,6 +9,9 @@ on:
     tags:
     - "**"
 
+permissions:
+  packages: write
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


szabozoltan69:
@sunu @batpad Recently all our GO infrastructure is in Azure, including the ifrcgoacr container registry.
Why don't we use this container registry instead of introducing a new one? We can have multiple repos in it.

sunu:
@szabozoltan69 We can keep using Azure container registry if you like. But I think the Github registry provides two slight advantages in terms of developer experience:
1. It provides better visibility - the published docker images and helm charts listing will be visible right next to the code in the repo like this:
image

2. When we roll out the playground cluster for folks to try out and deploy their apps, we also want them to be able to publish their own Docker images and Helm charts. Instead of using the ifrcgoacr registry, which would mean we have to give out login details or take on the job of building the images ourselves, it would be better if everyone could do this on their own. By using the GitHub registry for the go-api, we set a consistent example for how other apps in the go ecosystem should do things too.

szabozoltan69:
Ok, @sunu – I can accept this for playground (or try-it-out) purposes.

Issue #2091, opened 2024-04-04
Local Units and Health Care Data Integration 


davidmuchatiza:
**Statement of the Problem**
*The professional health services mapping team has been actively engaged in [globally mapping health sites](https://ifrcorg.sharepoint.com/:x:/r/sites/GOIVfordeveloperteam/Shared%20Documents/General/Geospatial/Professional%20Health%20Services/Data/Copy%20of%20Master%20Data_Professional%20Health%20Services%20Mapping.xlsx?d=w58db50c3c225441ea46824d4c6d1c8d9&csf=1&web=1&e=92Qiqb) affiliated with National Societies. The project has successfully collected detailed data from 175 National Societies. Simultaneously, the IFRC GO team has been working on incorporating local units for National Societies onto the GO platform, providing a centralized platform for National Societies to manage local unit data. The health sites identified by the Health Mapping Initiative are considered part of the broader local units' dataset, falling under one of the five categories of local units. However, there exists a disparity in data structure, with the health sites' mapping delving deep into attributes associated with health facilities while local units have adopted minimum viable data standards. With the conclusion of the health mapping initiative project, there is consideration for utilizing the GO platform to host this data, enabling users to interact with it on the frontend, which aligns more closely with the goals of the local units project.*

**Solution to Merge the Two Projects**
To integrate these two datasets, the team is proposing the following steps to be undertake:
[...]

PR #2107, opened 2024-04-15
Add data_source_id / visibility handling

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2091


local_units/management/commands/import_local_units_health.py
```diff
@@ -45,7 +44,9 @@ def handle(self, *args, **options):
 
                 country = country_name_id_map[row['COUNTRY'].lower()]
                 type = local_unit_id_map[int(row['TYPE CODE'])]
-                is_public = row['VISIBILITY'].lower() == 'public'
+                visibility = 3 if row['VISIBILITY'].lower() == 'public' else 1
+                data_source_id = row['DATA SOURCE ID']
+                health_id = row['DATA SOURCE ID']
```
k9845, line 49: Can't we just have a single field to store this information?

szabozoltan69: Good idea!
My only reason was: the better visibility on Admim page. But ok, I'll keep only the unavoidable foreign key reference then.

Issue #2118, opened 2024-04-26
Specify Data Retrieval Year in FDRS Databank API


samshara:
https://github.com/IFRCGo/go-web-app/issues/884, for point 4, we need the year information.

PR #2119, opened 2024-04-26
Add fdrs fetched date

k9845:
Addresses
- https://github.com/IFRCGo/go-api/issues/2118

## Changes
 - Added field `fdrs_data_fetched_year`

# NOTE: Please run the following command on staging server
`docker-compose exec serve bash  ./manage.py ingest_databank`

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


databank/models.py
```diff
@@ -535,6 +535,11 @@ class CountryOverview(models.Model):
         verbose_name=_('HDR GII'),
         null=True, blank=True
     )
+    fdrs_data_fetched_year = models.CharField(
+        verbose_name=_('FDRS Data Fetched Year'),
+        null=True, blank=True,
+        max_length=50
+    )
```
thenav56, line 542: Any reason why we are storing year in string?

databank/management/commands/sources/FDRS.py
```diff
@@ -120,11 +121,15 @@ def load(country, overview, fdrs_data):
     if country.iso is None or fdrs_data is None:
         return
 
+    get_max_year = lambda data: max(int(item['year']) for item in fdrs_data.values())
+    fdrs_data_fetched_year = get_max_year(fdrs_data)
```
thenav56, line 125: Any reason why we need lambda here instead of simple statement?

thenav56 (changes requested):

databank/models.py
k9845: We get year as string from data source and populating that here.

databank/management/commands/sources/FDRS.py
k9845: Just creating this to look clean.

k9845: Replaced with normal statement

samshara (approved):

PR #2131, opened 2024-05-04
Add JWT_PRIVATE_KEY_BASE64_ENCODED config

thenav56:
Address JWT value new line issue

## Change for azure pipeline
- Generate an encoded string using
  ```bash
  cat secret-key | base64 -w 0
  ```
- Set `JWT_PRIVATE_KEY_BASE64_ENCODED=<single-line-encoded-str>`

> NOTE: Valid JWT_PRIVATE_KEY_BASE64_ENCODED will replace JWT_PRIVATE_KEY


szabozoltan69 (commented):
Nice solution for hiding secret value in K8s.
Though this change is not related to the begin / key / end issue, is it?

thenav56:
> Nice solution for hiding secret value in K8s. Though this change is not related to the begin / key / end issue, is it?

This will also remove the \n from the string. Which will remove the new-line issue we are having in helm.

szabozoltan69:
Maybe also the public keys need some kind of encoding magic to avoid \n issue.

szabozoltan69 (approved):

PR #2141, opened 2024-05-10
Update nginx to allow OPTIONS method

thenav56:
Addresses Static icon load issue

## Changes

* Allow OPTIONS for static files
* Allow additional sentry headers

# Discussion
We should instead utilize azure blob storage for static files and define CORS there.


szabozoltan69 (approved):
Nice catch

Issue #2048, opened 2024-02-19
Clean Emergency page translations


tovari:
We would like to clean the emergencies languages and provide dynamic translations for the historic data.
The attached xlsx lists those emergency IDs that needs some batch action.
Actions to be done per worksheet:
Emerg (UI- En, summ. - Fr) - copy `Summary [en]`,` Title [en]` fields content to `summary [fr]` and `Title [fr]`, then change the `Entity original language` to French. Initiate translation. - 40 items

Emerg (UI- En, summ.- Es) - copy Summary [en], Title [en] fields content to summary [es] and Title [es], then change the Entity original language to Spanish. Initiate translation. - 104 items

Emerg (UI- Es, summ.- En) -  Initiate translation. - 424 items
[Summary_emergencies-to_be_translated-v3.xlsx](https://github.com/IFRCGo/go-api/files/14332287/Summary_emergencies-to_be_translated-v3.xlsx)

udaynwa:
Hey @szabozoltan69 . Here is a draft PR to run manual translation. Let me know if any additional information is required.
https://github.com/IFRCGo/go-api/pull/2149

szabozoltan69:
Some clarification and usage hints would be highly appreciated.

szabozoltan69:
Could you give some hints @thenav56 how to continue this task?

thenav56:
Hey @szabozoltan69, I hope this helps. Here is the pseudo code we will need for this:

```python
from lang.tasks import ModelTranslator
from main.translation import TRANSLATOR_SKIP_FIELD_NAME, TRANSLATOR_ORIGINAL_LANGUAGE_FIELD_NAME

from api.models import Event

def en_to_fr():
    pks = [...]  # List of primary keys -- Or read the content from an xlsx file instead?
    events = Event.objects.filter(
        pk__in=pks,
        **{TRANSLATOR_SKIP_FIELD_NAME: False},
    )

    print('Translating EN->FR')
    for event in events:
        assert getattr(event, TRANSLATOR_ORIGINAL_LANGUAGE_FIELD_NAME) == 'en'

        # TODO: Should we read the content from an xlsx file instead?
        event.name_fr = event.name_en
        event.summary_fr = event.summary_en
        ...

        # Clear the original English fields - Auto translator will only fill empty fields
        event.name_en = ''
        event.summary_en = ''
        ...

        # Change the original language to French
[...]

PR #2149, opened 2024-05-16
Manual translation trigger

thenav56:
### Addresses
- #2048

### Changes
- Added an option to trigger specific models for translation.

### Usages
```python
# Import the ModelTranslator class from the lang module
from lang.tasks import ModelTranslator

# Define a list of models to translate
models_to_translate = []

# Display character counts for the specified models
ModelTranslator.show_characters_counts(only_models=models_to_translate)

# Run the translation process for the specified models
ModelTranslator().run(batch_size=10, only_models=models_to_translate)
```

> **Note**: Additionally, there's another file available at [IFRCGo/go-api/lang/management/commands/translate_model.py](https://github.com/IFRCGo/go-api/blob/develop/lang/management/commands/translate_model.py), which does not allow specifying `models_to_translate` via stdin.

> **Batch_size**: How many instances to translate for each model. None will be all

### DIFF
--- a/lang/tasks.py
+++ b/lang/tasks.py
@@ -8,8 +8,6 @@ from modeltranslation import settings as mt_settings
 from django.apps import apps as django_apps
 from django.db.models.functions import Length
 from django.db import models
-from django.db.models import Sum
-from django.db.models import Q
 from django.conf import settings
 from functools import reduce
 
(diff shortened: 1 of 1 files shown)


thenav56:
Changes are already included in https://github.com/IFRCGo/go-api/pull/2156

Issue #2166, opened 2024-06-04
[PROD] Django4 Admin geometry fields


szabozoltan69:
## Issue
*Recently it is not possible to update geometry fields (centroid, boundary box) from Django Admin.*

In DJango3 there was also a hidden textarea for e.g. centroids, with id="id_centroid":
```
<a href="javascript:geodjango_centroid.clearFeatures()" data-jzz-gui-player="true">Delete all Features</a>
<textarea id="id_centroid" class="vWKTField required" cols="150" rows="10" name="centroid">POINT (2245128.33 5034984.47)</textarea>
<script>geodjango_centroid.init();</script>
```
– but in Django4 they appear as real textareas (with rich text options even) below the maps.
[This](https://porgeto.hu/tmp/dj3maps.gif) would be the normal appearance.

In Django4 this is the above quoted code part:
```
<a href="javascript:geodjango_centroid.clearFeatures()">Delete all Features</a>
<textarea id="id_centroid" class="vWKTField required" cols="150" rows="10" name="centroid" style="display: none;" aria-hidden="true"></textarea><div role="application" class="tox tox-tinymce" style="visibility: hidden; width: 1120px; height: 360px;" aria-disabled="false"><div class="tox-editor-container">...</div></div>
<script>geodjango_centroid.init();</script>
```

szabozoltan69:
A (local) workaround found: if we remove 
```
    'selector': 'textarea'
```
from the TINYMCE_DEFAULT_CONFIG setting, everything works fine (with maps). Unfortunately also the other rich text fields converts to simple oldschool HTML textareas.

szabozoltan69:
Maybe this way? https://stackoverflow.com/questions/35600829/how-to-have-all-tinymce-editors-as-standard-except-selected

PR #2167, opened 2024-06-04
Fix boundary box and centroid map edit

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2166


thenav56 (approved):
Nice solution :D

Issue #2169, opened 2024-06-10
CronJob to sync ALL Molnix positions


batpad:
Currently, we only fetch open positions from Molnix. We want to write a separate cronjob, maybe run once a day or so, that also syncs "historical" positions and makes sure everything is correctly updated in GO, for the case where someone edits a position that is already closed.

Creating this as a stub ticket - will add more details.

batpad:
cc @tovari @anamariaescobar

PR #2172, opened 2024-06-12
All Molnix positions instead of only /open

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2169

### DIFF
--- a/api/management/commands/sync_molnix_long.py
+++ b/api/management/commands/sync_molnix_long.py
@@ -1,0 +1,573 @@
+from dateutil import parser as date_parser
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from sentry_sdk.crons import monitor
+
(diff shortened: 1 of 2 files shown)


batpad:
@szabozoltan69 does this significantly increase the time it takes to process the Molnix records? i.e. if there are lots of historical records, then just maybe this makes it very slow and we need to reconsider. If it does not, then this looks good to me!

Do we have a good way to test this carefully on current production data or so to make sure it does not do anything unexpected?

Otherwise, this looks good to me!

szabozoltan69:
@batpad I don't know any sharp method for thorough testing. The run is not 10x longer than the old one, just 30-40% longer. It would be fine to "hunt" to a test case that is missing in the old workflow and could be found according to the new one.

tovari:
@szabozoltan69, @batpad I guess, we need to run this not more frequently than once a day

PR #2188, opened 2024-06-29
Fix event-pk query n+1 selects

szabozoltan69:
Refers to Sentry issue organizations/ifrc-go/issues/2073


szabozoltan69:
Before, after:
kép

PR #2228, opened 2024-07-25
Upgrade to python 3.11

szabozoltan69:
Python leverage


thenav56 (approved):
Looks good. Thanks, @szabozoltan69 

We'll need to check if the upgrade affects current dependencies using the staging environment.

szabozoltan69:
(I am a bit uncertain about is it really a good idea to upgrade, because the compile time was so much more than with Python 3.9... does 3.11 bring really so much goodies that it is worthy to do it?)

susilnem (approved):

Issue #1917, opened 2023-09-27
Allow read-only access to database for trusted users


batpad:
Part of https://github.com/IFRCGo/go-api/issues/1885

Sometimes. an IFRC user needs read-only access to the database for the purposes of data analysis, etc.

In the "old way", we setup an ssh tunnel VM that people could tunnel through and access the database, using user accounts with limited access.

With the new setup, we should figure out the best practice to do this.

It does seem to me like setting up an ssh jumpbox in the cluster and adding keys of trusted users makes sense.

@sunu would you have a recommendation on the best way to do this? And is this something you might be able to take on?

cc @szabozoltan69

sunu:
@batpad The only downside (that I can think of) in setting up a ssh jumpbox inside the k8s cluster instead of a VM outside the cluster is that users who can ssh into the jumpbox will also have access to internal services that are not exposed to the public internet by default. So we will have to apply a precise NetworkPolicy to prevent access to services other than the db in question from the jumpbox.

I can take this on. But I would need an environment to experiment safely and a bit of handholding when deploying the changes.

PR #2231, opened 2024-08-05
SSH bastion service

sunu:
Contributes towards https://github.com/IFRCGo/go-api/issues/1917

## Changes

* Creates a service that exposes a ssh server running inside the K8s cluster

### DIFF
--- a/deploy/helm/ifrcgo-helm/templates/bastion.yaml
+++ b/deploy/helm/ifrcgo-helm/templates/bastion.yaml
@@ -1,0 +1,99 @@
+{{- if .Values.sshBastion.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "ifrcgo-helm.fullname" . }}-ssh-authorized-keys
+data:
+{{- range $key, $value := .Values.sshBastion.keys }}
+  {{ $key }}: {{ $value | quote }}
+{{- end }}
+
+---
+
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ template "ifrcgo-helm.fullname" . }}-ssh-bastion-config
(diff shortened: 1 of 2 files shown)


batpad:
@sunu this looks really good to me!

Can we add some documentation for how this should work? Can just add it to this PR and @szabozoltan69 and @tovari can help us figure out where we should move it.

Taking a quick stab at the docs, based on our Slack conversation:

First, we need to setup a read-only user on the postgres database - @szabozoltan69 would be good to get your help in doing this.

Once that is setup:

 - Add user public keys to the `values` file like so: https://github.com/IFRCGo/go-api/pull/2231/files#diff-50ed908063f22cce00c15f9975df78704d8d52c8ef067e30a10612d1cf213692R186
 - Merge / Deploy this PR
 - It will create a Load Balancer resource, which will have its own IP address. We need to get the IP address for this new Load Balancer with something like `kubectl get svc`
 - Now, a user whose public key has been added can create an SSH tunnel and access the postgres database locally as the read-only user create above: `ssh -L {{local_port}}:{{database_host}}:{{remote_port}} user@{{ssh_host}}` to create the SSH tunnel. Once we have the values for these things, we will create better examples for this command. This will use SSH to expose the remote database on a local port.
 - To connect to the db, something like: `psql -h localhost -p {{local_port}} -U {{db_user}} -d {{db_name}}`

@szabozoltan69 whenever you're around, we can setup the database user, add some public keys, and test this out and once it all works, put some better docs and examples for the people who need to use this.

szabozoltan69:
In the past I introduced a read only postgresql user (Paola), but it needed a grant for every (important) tables. Maybe there is a better way than the one-by-one granting, because in this case every time a new table occurs we need to manually follow them.

Issue #2277, opened 2024-10-01
[PROD] Skip non-working endpoints in NS_INITIATIVES


szabozoltan69:
## Issue
Recently in https://github.com/IFRCGo/go-api/blob/develop/api/management/commands/ingest_ns_initiatives.py#L22-L24 we collect 3 data-api.ifrc.org endpoint results. If one of them is erroneous (like recently, when CBF gives error 500), the whole job fails due to the united results in all_fund_data.

The suggestion is: let's check the sanity of the results before uniting them.

szabozoltan69:
Deployed to Staging

PR #2278, opened 2024-10-01
Skip non-working ns_initiatives endpoints

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2277
Please check it, @susilnem – if you have a better solution, feel free to finalize it.

### DIFF
--- a/api/management/commands/ingest_ns_initiatives.py
+++ b/api/management/commands/ingest_ns_initiatives.py
@@ -19,19 +19,21 @@ class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         logger.info("Starting NS Inititatives")
         api_key = settings.NS_INITIATIVES_API_KEY
-        esf_url = requests.get(f"https://data-api.ifrc.org/api/esf?apikey={api_key}")
-        nsia_url = requests.get(f"https://data-api.ifrc.org/api/nsia?apikey={api_key}")
-        cbf_url = requests.get(f"https://data-api.ifrc.org/api/cbf?apikey={api_key}")
+        urls = [
+            f"https://data-api.ifrc.org/api/esf?apikey={api_key}",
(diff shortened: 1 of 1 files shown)


api/management/commands/ingest_ns_initiatives.py
```diff
@@ -24,9 +24,9 @@ def handle(self, *args, **kwargs):
         cbf_url = requests.get(f"https://data-api.ifrc.org/api/cbf?apikey={api_key}")
 
         # resposne for individual request
-        esf_response = esf_url.json()
-        nsia_response = nsia_url.json()
-        cbf_response = cbf_url.json()
+        esf_response = esf_url.json() if esf_url.status_code == 200 else []
+        nsia_response = nsia_url.json() if nsia_url.status_code == 200 else []
+        cbf_response = cbf_url.json() if cbf_url.status_code == 200 else []
```
susilnem, line 29:
```suggestion
   urls = [
            f"https://data-api.ifrc.org/api/esf?apikey={api_key}",
            f"https://data-api.ifrc.org/api/nsia?apikey={api_key}",
            f"https://data-api.ifrc.org/api/cbf?apikey={api_key}",
        ]

        responses = []
        for url in urls:
            response = requests.get(url)
            responses.append(response.json()) if response.status_code == 200 else responses.append([])

        added = 0
        flatList = [element for innerList in responses for element in innerList]
```

susilnem (dismissed):

api/management/commands/ingest_ns_initiatives.py
szabozoltan69: I think no "append nothing" is needed if the status_code is not 200.
(In the above "if"-s it was mandatory to give an "else" option.)

szabozoltan69: Thanks, I used your suggestions, the code became cleaner.

Issue #2293, opened 2024-10-25
[PROD] Data validation for OpsLearning while filling data to the db


szabozoltan69:
## Issue
For OpsLearning records there is a validation workflow process:
1. Data are not validated
2. Not validated data can be changed
3. Validation happens
4. Data are validated (and is_validated flag is true)
5. Validated data can be changed

If the is_validated field is false, we can suppose that the validation has not happened. For such records we should not allow to have ..._validated fields to be filled in (neither from Admin page, neither from API input).

szabozoltan69:
@thenav56 @susilnem @batpad ?

Tarekchehahde:
Opened PR: https://github.com/IFRCGo/go-api/pull/

**Local test summary:** Added API tests for reject/allow paths. Full Django suite not run locally — cc @szabozoltan69

PR #2294, opened 2024-10-26
Simplify OpsLearning Admin validation check

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2293

The OpsLearning Admin validation was not comfortable in case of API data was received previously.


szabozoltan69:
@ypyelab @tovari – fixing "API data disturbance" on OpsLearning Admin workflow.

Issue #2310, opened 2024-11-19
Argo hooks implementation


susilnem:
We can use Argo CD hooks to run actions after each deployment. Some possible actions include:
 
- Running Django migrations
- Collecting static files
- Creating static users for the alpha instances

> **Note:** Argo CD is currently being set up by @Lusengeri and is not yet available in the staging or production clusters. For now, we'll be using it in the alpha cluster, and once it's ready, we’ll move to staging and production.

cc: @szabozoltan69 

Reference: 
- https://kubernetes.io/docs/concepts/workloads/controllers/job/
- https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/

Lusengeri:
@susilnem , ArgoCD is available on Staging.

PR #2320, opened 2024-11-22
Add Argo hooks Implementation

susilnem:
Addresses
- https://github.com/IFRCGo/go-api/issues/2310

## Changes
- Add meta annotation for the celery.
- Fix config map for ELASTIC_SEARCH_HOST
- Add post sync argo hooks:
  - [x] migrations
  - [x] Collect static
  - [x] Alpha users

## Checklist

- [x] Updated/ran unit tests
- [x] Updated CHANGELOG.md

## Release

### DIFF
--- a/.github/workflows/build-publish-docker-helm.yml
+++ b/.github/workflows/build-publish-docker-helm.yml
@@ -3,6 +3,7 @@ name: Builds and pushes Docker Images and Helm charts to Github Registry
 on:
   # Build and push Docker image and Helm charts on every push to develop branch
   # and on every tag push
+  workflow_dispatch:
   push:
     branches:
     - develop
--- a/deploy/helm/ifrcgo-helm/.gitignore
+++ b/deploy/helm/ifrcgo-helm/.gitignore
@@ -1,0 +1,1 @@
+values-local.yaml
--- a/deploy/helm/ifrcgo-helm/.helmignore
+++ b/deploy/helm/ifrcgo-helm/.helmignore
@@ -1,0 +1,1 @@
+values-local.yaml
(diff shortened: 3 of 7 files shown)


szabozoltan69 (approved):

thenav56 (changes requested):

thenav56:
Let's test this using tc-local talos... then we can merge this

susilnem:
Hi @szabozoltan69 @Lusengeri,  
Could you please review this and merge if everything looks good?  
Thanks!

Lusengeri (commented):
At the moment, the go-api isn't being managed using argocd. As such, the included argo hooks will not have any effect. I would suggest we try this out on sandbox environment. Also, using argoCD would mean replacing the current deploy pipeline. We should test this out properly before replacing the current helm-release based deployment. I am available to support this change but we would need to coordinate.

thenav56:
> @susilnem , at the moment, the go-api isn't being managed using argocd. As such, the included argo hooks will not have any effect. I would suggest we try this out on sandbox environment first. Also, using argoCD would mean replacing the current deploy pipeline. We should test this out properly before replacing the current helm-release based deployment. I am available to support this change but we would need to coordinate.

Thanks @Lusengeri for the review!

We’re using argo-hooks for all alpha instances internally, which is a good starting point for testing and future ArgoCD integration in the go-api. It’s a good time to review and experiment with it. If you have any quick feedback on the implementation, please share!

@szabozoltan69 If you're interested, @susilnem can share a quick preview during the next dev call 😃

thenav56:
We have disabled the argo-hooks by default... it shouldn't impact the current deployment pipeline.

Issue #2333, opened 2024-12-05
Implement API for Operational Learning statistics


samshara:
This API must also support all filters applicable to the Operational Learning page.
The response should return the following fields:
- **`operations_included`**: Total count of operations based on the applied filters.
- **`sources_used`**: Count of unique sources referenced in the learning data.
- **`learning_extracts`**: Total number of validated learning extracts.
- **`sectors_covered`**: Count of distinct sectors covered.
- **`sources_overtime`**: Year-wise breakdown of sources used, categorized by type:
  - `DREF`
  - `Emergency Appeal`
  - `International Appeal`
  - `Forecast-Based Action (FBA)`
- **`learning_by_region`**: Learning data grouped by region:
  - `region_id`: Region ID
  - `region_name`: Region name
  - `count`: Number of learning extracts for the region.
- **`learning_by_sectors`**: Learning data grouped by sectors:
  - `sector_id`: Sector ID
  - `sector_name`: Sector name
  - `count`: Number of learning extracts for the sector.
- **`learning_by_country`**: Learning data for map visualization, grouped by country:
  - `country_id`: Country ID
  - `country_name`: Country name
  - `learning_count`: Number of learning extracts in the country.

PR #2335, opened 2024-12-09
Operational learning stats API

sudip-khanal:
Addresses 
- https://github.com/IFRCGo/go-api/issues/2333
## Changes

* Add ops-learning stats api 
* Add unit test 

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release

### DIFF
--- a/per/drf_views.py
+++ b/per/drf_views.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import pytz
 from django.conf import settings
 from django.db import transaction
-from django.db.models import Prefetch, Q
+from django.db.models import Count, F, OuterRef, Prefetch, Q, Subquery
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import get_language as django_get_language
@@ -20,7 +20,7 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.settings import api_settings
 
-from api.models import Country
+from api.models import Appeal, AppealType, Country
 from deployments.models import SectorTag
 from main.permissions import DenyGuestUserMutationPermission, DenyGuestUserPermission
 from main.utils import SpreadSheetContentNegotiation
@@ -920,6 +920,82 @@ class OpsLearningViewset(viewsets.ModelViewSet):
             )
         )
         return response.Response(OpsLearningSummarySerializer(ops_learning_summary_instance).data)
+
+    @action(
+        detail=False,
+        methods=["GET"],
+        permission_classes=[DenyGuestUserMutationPermission, OpsLearningPermission],
+        url_path="stats",
+    )
+    def stats(self, request):
+        """
+        Get the Ops Learning stats based on the filters
+        """
+        ops_data = (
+            super()
+            .get_queryset()
+            .filter(is_validated=True)
+            .select_related("appeal_code")
+            .prefetch_related(
+                "appeal_code__appealdocument",
+                "sector_validated",
+            )
+            .aggregate(
+                operations_included=Count("appeal_code", distinct=True),
+                learning_extracts=Count("id", distinct=True),
+                sector_covered=Count("sector_validated", distinct=True),
+                source_used=Count("appeal_code__appealdocument", distinct=True),
+            )
+        )
+
+        learning_by_sector = (
+            SectorTag.objects.filter(title__isnull=False)
+            .annotate(count=Count("validated_sectors", distinct=True))
+            .values("title", "count")
+        )
+
+        sources_overtime = {
+            str(appeal_type_label): OpsLearning.objects.filter(appeal_code__atype=appeal_type, is_validated=True)
+            .annotate(date=F(("appeal_code__start_date")))
+            .values("date")
+            .annotate(count=Count("appeal_code__appealdocument", distinct=True))
+            .order_by("date")
+            for appeal_type, appeal_type_label in AppealType.choices
+        }
+
+        region_subquery = Appeal.objects.filter(code=OuterRef("appeal_code"), region__isnull=False).values("region__label")[:1]
+        region_id = Appeal.objects.filter(code=OuterRef("appeal_code"), region__isnull=False).values("region__id")[:1]
+
+        learning_by_region = (
+            OpsLearning.objects.filter(is_validated=True)
+            .annotate(name=Subquery(region_subquery))
+            .values("name")
+            .annotate(id=Subquery(region_id), count=Count("id", distinct=True))
+            .order_by("name")
+        )
+
+        country_subquery = Appeal.objects.filter(code=OuterRef("appeal_code"), country__isnull=False).values("country__name")[:1]
+        country_id = Appeal.objects.filter(code=OuterRef("appeal_code"), country__isnull=False).values("country__id")[:1]
+
+        learning_by_country = (
+            OpsLearning.objects.filter(is_validated=True)
+            .annotate(name=Subquery(country_subquery))
+            .values("name")
+            .annotate(id=Subquery(country_id), count=Count("id", distinct=True))
+            .order_by("name")
+        )
+
+        data = {
+            "operations_included": ops_data["operations_included"],
+            "learning_extracts": ops_data["learning_extracts"],
+            "sectors_covered": ops_data["sector_covered"],
+            "sources_used": ops_data["source_used"],
+            "learning_by_region": learning_by_region,
+            "learning_by_sector": learning_by_sector,
+            "sources_overtime": sources_overtime,
+            "learning_by_country": learning_by_country,
+        }
+        return response.Response(data)
 
 
 class PerDocumentUploadViewSet(viewsets.ModelViewSet):
--- a/per/drf_views.py
+++ b/per/drf_views.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import pytz
 from django.conf import settings
 from django.db import transaction
-from django.db.models import Count, F, OuterRef, Prefetch, Q, Subquery
+from django.db.models import Count, F, Prefetch, Q
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import get_language as django_get_language
@@ -20,7 +20,7 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.settings import api_settings
 
-from api.models import Appeal, AppealType, Country
+from api.models import AppealType, Country, Region
 from deployments.models import SectorTag
 from main.permissions import DenyGuestUserMutationPermission, DenyGuestUserPermission
 from main.utils import SpreadSheetContentNegotiation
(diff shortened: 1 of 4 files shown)


per/test_views.py
```diff
@@ -224,3 +229,90 @@ def test_summary_generation(self, generate_summary):
...
+            self.assertIn(label, response.data["sources_overtime"])
+            for item in response.data["sources_overtime"][label]:
+                self.assertIn("year", item)
+                self.assertIn("count", item)
+
+        # Validate learning_by_region
+        for item in response.data["learning_by_region"]:
+            self.assertIn("region_name", item)
+            self.assertIn("region_id", item)
+            self.assertIn("count", item)
```
sudan45, line 294: compare the value instead of key

Rup-Narayan-Rajbanshi (approved):

per/drf_views.py
```diff
@@ -921,6 +922,65 @@ def summary(self, request):
...
+        permission_classes=[DenyGuestUserMutationPermission, OpsLearningPermission],
+        url_path="stats",
+    )
+    def stats(self, request):
+        """
+        Get the Ops Learning stats based on the filters
+        """
+        ops_data = OpsLearning.objects.aggregate(
+            operations_included=Count("learning", distinct=True),
+            learning_extracts=Count("learning_validated", distinct=True),
```
sudan45, line 937: don't count text field date

per/test_views.py
```diff
@@ -224,3 +229,90 @@ def test_summary_generation(self, generate_summary):
...
+
+        sectors_covered_count = (
+            OpsLearning.objects.filter(sector_validated__isnull=False).values("sector_validated").distinct().count()
+        )
+        self.assertEqual(response.data["sectors_covered"], sectors_covered_count)
+
+        # Validate sources_overtime
+        for appeal_type, label in AppealType.choices:
+            self.assertIn(label, response.data["sources_overtime"])
+            for item in response.data["sources_overtime"][label]:
```
sudan45, line 286: Compare values instead of keys

per/drf_views.py
sudan45, line 937: don't count text field data

sudan45 (changes requested):

susilnem (changes requested):
Let's Check the queries and remove duplicated queries.
mostly for country and region.

susilnem (changes requested):
Let's add some filter on it

per/test_views.py
```diff
@@ -224,3 +229,93 @@ def test_summary_generation(self, generate_summary):
...
+                self.assertIn("date", item)
+                self.assertIn("count", item)
+
+                date_str = item["date"]
+                date_str_iso = date_str.replace(tzinfo=None).isoformat() + "Z"
+
+                if label == "DREF":
+                    self.assertEqual(date_str_iso, "2023-01-01T00:00:00Z")
```
sudan45, line 316: compare date only 
```suggestion
                    self.assertEqual(date_str_iso, "2023-01-01")
```

per/drf_views.py
```diff
@@ -921,6 +921,82 @@ def summary(self, request):
...
+        sources_overtime = {
+            str(appeal_type_label): OpsLearning.objects.filter(appeal_code__atype=appeal_type, is_validated=True)
+            .annotate(date=F(("appeal_code__start_date")))
+            .values("date")
+            .annotate(count=Count("appeal_code__appealdocument", distinct=True))
+            .order_by("date")
+            for appeal_type, appeal_type_label in AppealType.choices
+        }
+
+        region_subquery = Appeal.objects.filter(code=OuterRef("appeal_code"), region__isnull=False).values("region__label")[:1]
```
sudan45, line 966: use select_related field

sudan45 (changes requested):

per/drf_views.py
```diff
@@ -921,6 +922,65 @@ def summary(self, request):
...
+        ops_data = queryset.aggregate(
+            operations_included=Count("appeal_code", distinct=True),
+            learning_extracts=Count("id", distinct=True),
+            sector_covered=Count("sector_validated", distinct=True),
+            source_used=Count("appeal_code__appealdocument", distinct=True),
+        )
+
+        learning_by_sector = (
+            SectorTag.objects.filter(title__isnull=False)
+            .annotate(count=Count("validated_sectors", distinct=True))
```
sudan45, line 950: use  filterset

sudan45 (changes requested):

sudan45 (approved):

per/serializers.py
```diff
@@ -1253,3 +1253,36 @@ class Meta:
...
+    count = serializers.IntegerField()
+
+
+class LearningBySectorSerializer(serializers.Serializer):
+    title = serializers.CharField()
+    count = serializers.IntegerField()
+
+
+class LearningSourcesOvertimeSerializer(serializers.Serializer):
+    type = serializers.IntegerField()
```
Rup-Narayan-Rajbanshi, line 1275: @susilnem Do we return type as integer or get display name for type? If no may be we need to do `type = serializers.CharField(source='get_type_display', read_only=True)` to get the type name.

susilnem: No, for now it's not required.

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+            operations_included=Count("appeal_code", distinct=True),
+            learning_extracts=Count("id", distinct=True),
+            sector_covered=Count("sector_validated", distinct=True),
+            source_used=Count("appeal_code__appealdocument", distinct=True),
+        )
+
+        learning_by_sector = (
+            SectorTag.objects.filter(validated_sectors__in=queryset, title__isnull=False)
+            .annotate(count=Count("validated_sectors", distinct=True))
+            .values("title", "count")
```
thenav56, line 951: We also need to add 'id', this will be used by client as unique key

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+        )
+
+        sources_overtime = (
+            Appeal.objects.filter(opslearning__in=queryset)
+            .annotate(
+                type=F("atype"),
+                date=F("start_date"),
+                count=Count("appealdocument", distinct=True),
+            )
+            .values("type", "date", "count")
```
thenav56, line 961: We also need to add 'id', this will be used by client as unique key

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+        """
+        queryset = self.filter_queryset(self.get_queryset()).filter(is_validated=True)
+        ops_data = queryset.aggregate(
+            operations_included=Count("appeal_code", distinct=True),
+            learning_extracts=Count("id", distinct=True),
+            sector_covered=Count("sector_validated", distinct=True),
+            source_used=Count("appeal_code__appealdocument", distinct=True),
+        )
+
+        learning_by_sector = (
```
thenav56, line 948: Let's add `_qs` as postfix (`learning_by_sector_qs`) for all querysets variables

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+            .values("type", "date", "count")
+        )
+
+        learning_by_region = (
+            Region.objects.filter(appeal__opslearning__in=queryset)
+            .annotate(
+                region_name=F("label"),
+                count=Count("appeal__opslearning", distinct=True),
+            )
+            .values("region_name", "count")
```
thenav56, line 970: We also need to add 'id', this will be used by client as unique key

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+        )
+
+        learning_by_country = (
+            Country.objects.filter(appeal__opslearning__in=queryset)
+            .annotate(
+                country_id=F("id"),
+                country_name=F("name"),
+                count=Count("appeal__opslearning", distinct=True),
+            )
+            .values("country_id", "country_name", "count")
```
thenav56, line 980: We also need to add 'id', this will be used by client as unique key

per/drf_views.py
```diff
@@ -921,6 +922,76 @@ def summary(self, request):
...
+            source_used=Count("appeal_code__appealdocument", distinct=True),
+        )
+
+        learning_by_sector = (
+            SectorTag.objects.filter(validated_sectors__in=queryset, title__isnull=False)
+            .annotate(count=Count("validated_sectors", distinct=True))
+            .values("title", "count")
+        )
+
+        sources_overtime = (
```
thenav56, line 954: Let's add a NOTE that this queryset is unbounded, and we may need to add some start_date filter in the future.

per/serializers.py
thenav56: Are we sure about this? It will be complicated for the client side to do this mapping.

`source='get_type_display'` can't be used here, as the object will be a dict instead of a proper object. We could use serializer_method or do the enum->label in the views.

per/factories.py
```diff
@@ -105,12 +107,22 @@ class Meta:
         model = FormPrioritization
 
 
+class AppealFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = Appeal
+
+    country = factory.SubFactory(CountryFactory)
```
thenav56, line 114: Let's not use `factory.SubFactory`, this will create objects without developer knowledge which can have weird confusing related to data generation, and it's effect.

per/factories.py
```diff
@@ -105,12 +107,22 @@ class Meta:
...
+
+
 class OpsLearningFactory(factory.django.DjangoModelFactory):
     learning = fuzzy.FuzzyText(length=50)
 
     class Meta:
         model = OpsLearning
 
+    appeal_code = factory.SubFactory(AppealFactory)
+    is_validated = fuzzy.FuzzyChoice([True, False])
```
thenav56, line 124: Let's not use fuzzy choice for is_validated, we should manually define in the test cases as lots of endpoints depends on this

per/serializers.py
```diff
@@ -1253,3 +1254,36 @@ class Meta:
...
+
+class OpsLearningStatSerializer(serializers.Serializer):
+    operations_included = serializers.IntegerField()
+    learning_extracts = serializers.IntegerField()
+    sectors_covered = serializers.IntegerField()
+    sources_used = serializers.IntegerField()
+    learning_by_region = LearningByRegionSerializer(many=True)
+    learning_by_country = LearningByCountrySerializer(many=True)
+    learning_by_sector = LearningBySectorSerializer(many=True)
+    sources_overtime = LearningSourcesOvertimeSerializer(many=True)
```
thenav56, line 1289: Let's explicitly define required=True for all fields

per/test_views.py
```diff
@@ -224,3 +229,72 @@ def test_summary_generation(self, generate_summary):
...
+        self.ops_learning2.sector_validated.set([self.sector2])
+
+        self.ops_learning3 = OpsLearningFactory.create(is_validated=False, appeal_code=self.appeal2)
+        self.ops_learning3.sector_validated.set([self.sector2])
+
+    def test_ops_learning_stats(self):
+        url = "/api/v2/ops-learning/stats/"
+        response = self.client.get(url)
+
+        self.assert_200(response)
```
thenav56, line 267: If we are comparing all fields data, it will be better to just use basic dict assert instead.

`assert response.data == expected_data`

thenav56 (changes requested):
Looks good. Few integration required

PR #2355, opened 2024-12-26
Generate title api for field report

susilnem:
Addresses
- Api for generating title sample before creating field report

## Changes
- Add api for field report title generating

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


api/drf_views.py
```diff
@@ -943,6 +945,41 @@ def get_serializer_class(self):
...
+        summary = generate_field_report_title(
+            country=countries[0], dtype=dtype, event=event, start_date=start_date, title=title, is_covid_report=is_covid_report
+        )
+        return Response(
+            FieldReportGeneratedTitleSerializer(
+                {
+                    "title": summary,
+                }
+            ).data
+        )
```
Rup-Narayan-Rajbanshi, line 981: Why use serializer for a single field. Can't we return  `return Response{'title': summary}`

susilnem: using the serializer for typing

api/utils.py
```diff
@@ -96,3 +98,37 @@ def write(self, value):
 
 def bad_request(message):
     return JsonResponse({"statusCode": 400, "error_message": message}, status=400)
+
+
+def generate_field_report_title(
+    country: models.Model,
+    dtype: models.Model,
+    event: models.Model,
```
thenav56, line 106: We should use proper model instead of base class models.Model

If there is import error, we can use
```
if typing.TYPE_CHECKING:
    from api.models import Country


....

    country: 'Country',
```

api/utils.py
```diff
@@ -96,3 +98,37 @@ def write(self, value):
...
+    start_date: timezone.datetime,
```
thenav56, line 107: Is `start_date` optional?

thenav56 (changes requested):
Looks good

api/utils.py
susilnem: yes

thenav56 (approved):

PR #2358, opened 2024-12-27
django-oauth-toolkit setup

thenav56:


szabozoltan69:
I am so glad that this topic is under development!

.python-version
```diff
@@ -1 +1 @@
-3.11.10
```
tnagorra, line 1: Not sure if we need to change this version. There was an issue during setup but do not remember if this change fixed the issue.

api/templates/login.html
```diff
@@ -0,0 +1,68 @@
...
+    font-size: 0.875rem;
+}
+
+form input[type="password"],
+form input[type="text"] {
+    border: unset;
+    background-color: rgba(0, 0, 0, 0.1);
+    padding: 0.5rem;
+    border-radius: 0.25rem;
+}
```
tnagorra, line 30: We might be able to move this to base styles.

main/settings.py
```diff
@@ -700,6 +712,31 @@ def decode_base64(env_key, fallback_env_key):
 AZURE_OPENAI_KEY = env("AZURE_OPENAI_KEY")
 AZURE_OPENAI_DEPLOYMENT_NAME = env("AZURE_OPENAI_DEPLOYMENT_NAME")
 
+# FIXME: Do not hard-code http protocol
+LOGIN_REDIRECT_URL = f"http://{FRONTEND_URL}/permalink/login-callback"
+LOGOUT_REDIRECT_URL = f"http://{FRONTEND_URL}/permalink/logout-callback"
```
tnagorra, line 717: Do we need to define this? We have not added any change to go-web-app regarding the SSO login

main/urls.py
```diff
@@ -169,6 +172,9 @@
 admin.site.site_title = "IFRC Go admin"
 
 urlpatterns = [
+    path("login/", LoginFormView.as_view(), name="go_login"),
+    path("logout/", logout_user, name="go_logout"),
```
tnagorra, line 176: We should rename the url names. Maybe /sso-auth/ and /sso-logout

main/settings.py
thenav56: Using `/sso-auth/` for the redirects for now

main/oauth2.py
```diff
@@ -0,0 +1,18 @@
+from django.contrib.auth.models import User
+from oauth2_provider.oauth2_validators import OAuth2Validator
+
+
+class CustomOAuth2Validator(OAuth2Validator):
+
+    def get_additional_claims(self, request):
+        user: User = request.user
+        user.get_full_name()
```
tnagorra, line 9: Do we need to call this function?

thenav56: Ops.. removed

tnagorra (approved):

Issue #2362, opened 2024-12-30
Remove component 14 sub components from PER component filter


susilnem:
## Migrate Sub-Components to Component 14

- Remove component 14 sub-components from the PER filter and admin panel.
- Migrate data tagged as sub-components of component 14 to component 14.

PR #2372, opened 2025-01-05
Add logic to remove subcomponent of component14

sudip-khanal:
Addresses 
* #2362

## Changes
* Add migration to migrate sub-component of component 14 to component 14
* Remove sub-component of component 14 form admin and filter

## Checklist

- [x] Updated/ran unit tests
- [x] Updated CHANGELOG.md

## Release


per/drf_views.py
```diff
@@ -175,6 +175,12 @@ class Meta:
         model = FormComponent
         fields = {"area": ("exact",)}
 
+    def filter_queryset(self, queryset):
+        queryset = super().filter_queryset(queryset)
+
+        queryset = queryset.exclude(component_num=14, component_letter__isnull=False)
+        return queryset
+
```
susilnem, line 183: We should add a filter to exclude sub component...

susilnem:
```
    # Note: Custom filter to exclude subcomponents of component 14
    exclude_subcomponents = filters.BooleanFilter(
        help_text="Remove sub components of component 14",
        method="get_exclude_subcomponents",
    )
    
    def get_exclude_subcomponents(self, queryset, name, value):
        if value:
            return queryset.exclude(component_num=14, is_parent__isnull=True)
        return queryset
```

per/migrations/0124_migrate_sub_component_to_component14.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+
+def migrate_sub_components_to_component14(apps, schema_editor):
+    FormComponent = apps.get_model("per", "FormComponent")
+    OpsLearning = apps.get_model("per", "OpsLearning")
+
+    component_14 = FormComponent.objects.filter(component_num=14)
```
susilnem, line 8:
```suggestion
    parent_component_14 = FormComponent.objects.filter(component_num=14, is_parent=True)
    sub_components_14 = FormComponent.objects.filter(component_num=14, is_parent__isnull=True)
```

per/migrations/0124_migrate_sub_component_to_component14.py
```diff
@@ -0,0 +1,26 @@
...
+    if component_14.exists():
```
susilnem, line 9:
```suggestion
    if not sub_components_14.exists():
            self.stdout.write(self.style.WARNING("No sub-components found for component 14."))
            return
```

per/migrations/0124_migrate_sub_component_to_component14.py
```diff
@@ -0,0 +1,26 @@
...
+        sub_components_14 = component_14.filter(component_letter__isnull=False)
+        ops_learnings = OpsLearning.objects.filter(per_component__in=sub_components_14)
+        for ops_learning in ops_learnings:
+            if ops_learning.is_validated:
+                ops_learning.per_component_validated.remove(*sub_components_14)
```
susilnem, line 14: We also need to check if the per_component_validated contains the sub component instead of just removing just like per_component .....

susilnem (changes requested):
Minor changes...
Let's just use a script instead of a migration file here.

Rup-Narayan-Rajbanshi (approved):

per/management/commands/migrate_sub_components_to_component14.py
```diff
@@ -0,0 +1,32 @@
...
+        sub_components_14 = FormComponent.objects.filter(component_num=14, is_parent__isnull=True)
+
+        if not sub_components_14.exists():
+            self.stdout.write(self.style.WARNING("No sub components found for component 14"))
+
+            return
+
+        sub_components_14_ids = sub_components_14.values_list("id", flat=True)
+
+        ops_learnings = OpsLearning.objects.filter(
```
thenav56, line 21:
```suggestion
        ops_learnings_qs = OpsLearning.objects.filter(
```

per/management/commands/migrate_sub_components_to_component14.py
```diff
@@ -0,0 +1,32 @@
...
+            per_component__in=sub_components_14_ids, per_component_validated__in=sub_components_14_ids
+        )
```
thenav56, line 23: Do we need iterator (Not required if we don't use for loop)

per/management/commands/migrate_sub_components_to_component14.py
```diff
@@ -0,0 +1,32 @@
...
+        for ops_learning in ops_learnings:
+            if ops_learning.is_validated:
+                ops_learning.per_component_validated.remove(*sub_components_14_ids)
+                ops_learning.per_component_validated.add(parent_component_14)
+            ops_learning.per_component.remove(*sub_components_14_ids)
+            ops_learning.per_component.add(parent_component_14)
```
thenav56, line 29: Maybe we can do this instead?
```suggestion
    # Change sub_component
    OpsLearning.per_component_validated.through.filter(ops_learning__is_validated=True).update(
        componenet=parent_component_14,
    )
    OpsLearning.per_component.through.update(
        componenet=parent_component_14,
    )
```

thenav56 (changes requested):

thenav56 (approved):
Let's add some noise data in the test case.

Let's QA using alpha instances

PR #2420, opened 2025-02-20
Feature/s3 storage

thenav56:
Addresses **PDF Export support for local and Alpha environments**

## Changes

* Allow using locally running go-web-app for PDF export
* Add support for S3 storage (using self-host Bitnami [minio](https://min.io/) sub-chart) to serve static/media files in Alpha environments where Azure blob storage is not available.
  * Eg: Export DREF using https://alpha-4.ifrc-go.dev.datafriendlyspace.org/account/my-forms/dref
* Add basic Bitnami Postgres sub-chart inside cluster for Alpha environments.
* Add config to remove hardcoded `https` in the system (http is required for local development) - WIP

## 🚨 Breaking change 🚨 
- We need to explicitly enable Azure blob in go-deploy using ENV `AZURE_STORAGE_ENABLED=true`

### DIFF
--- a/.github/workflows/build-publish-docker-helm.yml
+++ b/.github/workflows/build-publish-docker-helm.yml
@@ -42,7 +42,7 @@ jobs:
     - name: 🐳 Helm dependency
       working-directory: deploy/helm/ifrcgo-helm
       run: |
-        yq --indent 0 '.dependencies | map(["helm", "repo", "add", .name, .repository] | join(" ")) | .[]' ./helm/Chart.lock  | sh --
+        yq --indent 0 '.dependencies | map(["helm", "repo", "add", .name, .repository] | join(" ")) | .[]' ./Chart.lock | sh --
         helm dependency build ./
 
     - name: Tag docker image in Helm Chart values.yaml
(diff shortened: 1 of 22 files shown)


szabozoltan69:
I added to go-deploy CD variables AZURE_STORAGE_ENABLED as "true".

thenav56:
Hey @szabozoltan69, thanks for looking into this.

I have set `AZURE_STORAGE_ENABLED=true` as fallback in helm chart configmap, which will now handle the azure blog config automatically for now. https://github.com/IFRCGo/go-api/blob/a4dbbcae4ac068f1e07469984983d35e482a0baa/deploy/helm/ifrcgo-helm/templates/config/configmap.yaml#L19-L23

susilnem (approved):

szabozoltan69:
Thanks @thenav56 – so I can remove it from the CD variable list for now, not to "spam" it with unused ones.

szabozoltan69 (approved):
It's so great that we can use bitnami constructions!

thenav56:
Thanks, @szabozoltan69 
Sounds good on the clean-up.

We will need to do a brief test for the azure blob storage in staging to make sure these changes doesn't break anything.
cc: @udaynwa @shraeyaa

szabozoltan69:
Ok, I deploy develop branch as it gets ready with CI steps.

PR #2515, opened 2025-07-07
feat(local-unit): add model for externally managed local unit

sudip-khanal:
Addresses 
- https://github.com/orgs/IFRCGo/projects/11/views/12?pane=issue&itemId=118693423&issue=IFRCGo%7Cgo-api%7C2510
- https://github.com/orgs/IFRCGo/projects/11/views/12?pane=issue&itemId=118693354&issue=IFRCGo%7Cgo-api%7C2508

## Changes
- Add model for externally managed local unit
- Add api for externally managed local unit
- Add a field on local unit model for provide reason for the change after editing
## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


susilnem (changes requested):

local_units/serializers.py
```diff
@@ -590,3 +591,32 @@ def update(self, instance, validated_data):
...
+            "country",
+            "local_unit_type",
+            "enabled",
+            "country_details",
+            "local_unit_type_details",
+        )
+
+    def create(self, validated_data):
+        validated_data["created_by"] = self.context["request"].user
+        validated_data["updated_by"] = self.context["request"].user
```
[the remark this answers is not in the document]
sudip-khanal: I think created_by is needed here to track who created record.
We can skip updated by here

local_units/permissions.py
```diff
@@ -36,3 +37,12 @@ def has_permission(self, request, view):
         if request.method in ["POST", "PUT", "PATCH"]:
             return request.user and request.user.is_authenticated
         return True
+
+
+class ExternallyManagedLocalUnitPermission(permissions.BasePermission):
```
susilnem, line 42: Here, the external management is only managed by superuser?

If yes, Don't we need to set the permissions for other endpoints for safe-methods too for superuser only?

local_units/views.py
```diff
@@ -367,3 +374,13 @@ class DelegationOfficeDetailAPIView(RetrieveAPIView):
         permissions.IsAuthenticated,
         DenyGuestUserPermission,
     ]
+
+
+class ExternallyManagedLocalUnitViewSet(viewsets.ModelViewSet):
```
susilnem, line 379: Use generics and mixins here for C-U-R!

PR #2527, opened 2025-08-05
Add table_field to translation cache for stats

szabozoltan69:
To track what is translated


lang/translation.py
```diff
@@ -143,13 +143,17 @@ def translate_text(self, text, dest_language, source_language=None):
                 dest_language=dest_language,
             ).first()
             if cache:
-                logger.info(f"IFRC translation cache hit: {text[:30]}... {source_language}>{dest_language}")
+                cache.num_calls += 1
+                cache.save(update_fields=["num_calls"])
```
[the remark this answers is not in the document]
szabozoltan69: Done

lang/models.py
```diff
@@ -40,6 +40,8 @@ class TranslationCache(models.Model):
     source_language = models.CharField(max_length=16)
     dest_language = models.CharField(max_length=16)
     translated_text = models.TextField()
+    table_field = models.CharField(max_length=128, blank=True, default="")  # for stats only
```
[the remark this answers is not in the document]
szabozoltan69: It's clear that we should know if the recorded table_field is the only one that we use or not (for that text), but by my opinion it would make this workflow significantly slower if we would have a task to check a many-to-many field content. So a simple boolean field can be a (short term) solution to see if a record is used for one table_field or more. (If there are a lot of True-s, let's consider to bump up the structure with a *-to-many field.)

lang/models.py
```diff
@@ -37,16 +37,21 @@ def get_user_permissions_per_language(cls, user):
 
 class TranslationCache(models.Model):
     text = models.TextField()
+    text_hash = models.CharField(max_length=64)
```
thenav56, line 40: We should make this unique.

lang/migrations/0009_remove_translationcache_lang_transl_text_4a497b_idx_and_more.py
```diff
@@ -0,0 +1,52 @@
...
+from django.db import migrations, models
+
+
+def fill_text_hash(apps, schema_editor):
+    TranslationCache = apps.get_model("lang", "TranslationCache")
+    for obj in TranslationCache.objects.all():
+        if not obj.text_hash:
+            obj.text_hash = hashlib.sha256(obj.text.encode("utf-8")).hexdigest()
+            obj.save(update_fields=["text_hash"])
+    print(" Hash field is populated in lang_translationcache.")
```
thenav56, line 14: I recommend doing this in Postgres if possible.

https://www.postgresql.org/docs/current/pgcrypto.html

thenav56 (changes requested):
Thanks @szabozoltan69 for adding the hash support

lang/models.py
szabozoltan69: This will not be unique unless we include the source_language and dest_language also into the hash. (Otherwise normally we'll have 3 records with the same hash, with en>es, en>fr, en>ar.

Issue #2633, opened 2026-01-16
Molnix cronjob for Appraisals and Response Capacity


tovari:
We need a new cronjob or to extend the existing one to import appraisals and response capacity information from RRMS.

Technical details of the api endpoints are unclear, need to explore or get help from the Molnix team.

arunissun:
@szabozoltan69 and @tovari   
Hi Zoltan, I would request you if you can check the requirements. Let me know if you dont understand anything or something is wrong, thanks 

# Molnix RRMS Data Requirements and Simple Schema Proposal (4 Tables)

## 1) What This Document Does

This proposal is intentionally simple and aligned with your guidance:

1. only 4 new tables
2. person and event tables are RRMS-level (not deployments-prefixed)
3. appraisal tables are deployments-prefixed
4. all required fields are mapped
5. strict no personal info

## 2) Verified Facts (Molnix API + Postgres data)

1. Appraisals API is paginated.
2. `page=2` returns valid data.
3. Current appraisals response includes pagination fields: `current_page`, `last_page`, `next_page_url`, `per_page`, `total`.
4. In sampled appraisals, `appraisal.target_id` and `appraisal.deployment.id` are equal.
5. Appraisal deployment IDs do not match `deployments_personneldeployment.id` in DB sample checks.
6. `person_id` is an RRMS API person id when coming from people endpoints.
7. `personnel_id` is local GO DB key and must be resolved during ingest.
8. Events API is paginated in the same way as Appraisals API.
9. Organizations endpoint (`https://rrms.ifrc.org/api/api/system/organizations`) provides organization master data (`id` and `name`) for enrichment.

## 3) ID Clarification in Simple Words

### 3.1 person_id

1. comes from Molnix RRMS API
2. identifies a person in RRMS
3. used for RRMS-wide people data (deployed + not deployed or standby)

### 3.2 personnel_id

1. does not come directly from Molnix API
2. local key in the IFRC GO DB: `deployments_personnel.deployedperson_ptr_id`
3. filled only after matching RRMS data to local records

## 4) The 4 New Tables

## 4.1 rrms_person_snapshot

Why this table is needed:

1. dashboard needs total RRMS members and standby members
2. dashboard needs breakdowns by role and language
3. this table covers all RRMS persons (deployed and non-deployed or stand by )
4. We also need information on the tags becasue tags help us identify if the person registered on the RRMS is a member or not or on stand by  (simply registering does not mean a member)

Primary key:

[...]

PR #2670, opened 2026-02-25
Sync Molnix appraisals and appraisers

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/2633 and to https://github.com/IFRCGo/go-api/issues/2732


arunissun:
@szabozoltan69   
In api/management/commands/sync_molnix_appraisals.py, please check the block around lines 559-615. It looks like for appraisal in appraisals: starts correctly, but the actual normalize_appraisal(...) write logic and the nested appraiser handling are outside that loop, so the code appears to process only the last appraisal on each page and only that last appraisal’s appraisers. The ticket expects one row per appraisal and the full child set from appraisal.appraisers[], so could you please verify the indentation/control flow, 
let me know if I am missing something

szabozoltan69:
> ...please check the block around lines 559-615

Good one, @arunissun – I fixed the loop indentation and pushed the change.

arunissun:
@szabozoltan69 
Please could you confirm whether duplicate rrms_event_participation rows on rerun are expected by design? In sync_molnix_appraisals.py, the event participation records are written with create(), and the uniqueness constraint was removed in migration 0095_molnix_appraisal_appraised_person_and_event_participation_constraint.py. So if the sync is run again on the same source data, it looks like the same (event_id, person_id, role) participation row could be inserted again instead of being updated or skipped. If that behavior is intentional, then this point is fine from my side or I am missing anything , 

no more issues identified and you can merge the PR

szabozoltan69:
@arunissun You were right about duplicates. I solved it via this [commit](https://github.com/IFRCGo/go-api/pull/2670/commits/4c5b9e07953e4be67721cf97742c3eae0c85a14e): "Update RRMS event participation to upsert and log create/update counts".

PR #2711, opened 2026-04-09
chore(filter): change the datetime filter.

sandeshit:
## Changes

* Revert datetime filter to use start/end.
* Changed gdacs flood transformer to use sum of impacts.

## Checklist

- [x] Updated/ran unit tests
- [x] Updated CHANGELOG.md

### DIFF
--- a/alert_system/etl/gdacs_flood/transform.py
+++ b/alert_system/etl/gdacs_flood/transform.py
@@ -13,14 +13,14 @@ class GdacsTransformer(BaseTransformerClass):
     Extracts and normalizes impact fields, computes derived values, and stores metadata.
     """
 
-    # NOTE: This logic might change in future
+    # NOTE: This logic might change in future. Currently only creating a function for future use.
     def compute_people_exposed(self, metadata_list) -> Optional[int]:
         for data in metadata_list:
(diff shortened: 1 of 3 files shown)


alert_system/etl/gdacs_flood/transform.py
```diff
@@ -30,28 +30,35 @@ def compute_buildings_exposed(self, metadata_list) -> Optional[int]:
...
-                        "value": value,
-                        "unit": impact_detail.get("unit", ""),
-                        "estimate_type": impact_detail.get("estimate_type", ""),
-                    }
-                ]
+            detail = properties.get("monty:impact_detail", {})
+
+            category = detail.get("category")
+            type_ = detail.get("type")
+            value = detail.get("value") or 0
```
ranjan-stha: @sandeshit  In pystac-monty side, "0" carries a meaning so when the impact value information is not available, in few of the sources, the "value" might be set to -1 to indicate that there is no information on value (value being int so used -1). So, please control it here as well. If detail.get("value") is -1 also, set it to 0 as you are only calculating the sum per category and type_ basis.

ranjan-stha (approved):

PR #2717, opened 2026-04-18
Upgrade from Django 4 to Django 5.0 - 5.1 - 5.2

szabozoltan69:
Django 4 support ends in April.
Let's shift to the next LTS: https://static.djangoproject.com/img/release-roadmap.f1c2fefeeb00.svg


pyproject.toml
```diff
@@ -4,34 +4,32 @@ version = "1.1.419"
 description = ""
 authors = [{ name = "GO Dev", email = "go-dev@ifrc.org" }]
 license = { text = "MIT License" }
-requires-python = ">=3.11"
+requires-python = ">=3.11,<3.13"
```
thenav56: We should pin to a version

```suggestion
requires-python = "~=3.13.13"
```

We also need to update `.python-version` with
```
~=3.13.13
```

https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

api/management/commands/index_and_notify.py
```diff
@@ -1,12 +1,13 @@
...
-from datetime import datetime, timedelta, timezone
+from datetime import timedelta
 
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
 from django.db.models import DurationField, ExpressionWrapper, F, Q, Sum
 from django.db.models.query import QuerySet
 from django.template.loader import render_to_string
+from django.utils import timezone as dj_timezone
```
thenav56: Is there a reason we're using `dj_timezone` instead of just `timezone`?
Using `timezone` directly would avoid a lot of changes and confusion.

Also, I noticed `dt_timezone` is used in other places.

thenav56 (changes requested):
Looks good. Few comments from my side.

How are we planning to test this out?

pyproject.toml
szabozoltan69: The reason of dt_datetime usage was to differentiate this from django datatime. But where there is only 1, we can simplify the name, I agree. Also agree with the fixed python version.

api/management/commands/index_and_notify.py
szabozoltan69: dj_ means Django, dt_ means datetime.

PR #2731, opened 2026-04-30
Fix N+1 translation cache lookups in celery translation tasks

szabozoltan69:
For Sentry issue /organizations/ifrc-go/issues/5764

### DIFF
--- a/lang/tasks.py
+++ b/lang/tasks.py
@@ -39,23 +39,41 @@ class ModelTranslator:
         if not initial_value or not initial_lang:
             return
 
+        model = type(obj)
+        app_label = model._meta.app_label
+        model_name = model._meta.model_name
+        table_field = f"{app_label}:{model_name}:{field}"
+        pending_langs = []
+        for lang in AVAILABLE_LANGUAGES:
+            lang_field = build_localized_fieldname(field, lang)
+            if getattr(obj, lang_field, None):
+                continue
+            pending_langs.append(lang)
+
+        cached_translations = {}
+        if pending_langs:
+            cached_translations = self.translator.get_cached_translations(
+                initial_value,
+                pending_langs,
+                source_language=initial_lang,
+                table_field=table_field,
+            )
+
         for lang in AVAILABLE_LANGUAGES:
             lang_field = build_localized_fieldname(field, lang)
             value = getattr(obj, lang_field, None)
             if value:
                 continue
 
-            model = type(obj)
-            app_label = model._meta.app_label
-            model_name = model._meta.model_name
-            table_field = f"{app_label}:{model_name}:{field}"
-
-            new_value = self.translator.translate_text(
-                initial_value,
-                lang,
-                source_language=initial_lang,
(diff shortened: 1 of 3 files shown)


lang/tasks.py
thenav56, line 1: Maybe we can refactor the whole function with the new cache logic.
Something like this
```python
def translate_fields_object(self, obj, field):
    initial_lang = getattr(obj, TRANSLATOR_ORIGINAL_LANGUAGE_FIELD_NAME)
    initial_value = getattr(obj, build_localized_fieldname(field, initial_lang), None)
    if not initial_value or not initial_lang:
        return

    model = type(obj)
    table_field = f"{model._meta.app_label}:{model._meta.model_name}:{field}"
    field_max_length = model._meta.get_field(field).max_length

    # Determine which languages still need translation
    pending_langs = {
        lang: build_localized_fieldname(field, lang)
        for lang in AVAILABLE_LANGUAGES
        if not getattr(obj, build_localized_fieldname(field, lang), None)
    }
    if not pending_langs:
        return

    # Batch-fetch cached translations for all pending languages at once
    cached = self.translator.get_cached_translations(
        initial_value,
        list(pending_langs.keys()),
        source_language=initial_lang,
        table_field=table_field,
    )

    for lang, lang_field in pending_langs.items():
        translated = cached.get(lang) or self.translator.translate_text(
            initial_value,
            lang,
            source_language=initial_lang,
            table_field=table_field,
        )

        if translated is None:
            logger.warning(
                "Translation failed for %s.%s pk=%s",
                model.__name__,
                lang_field,
                obj.pk,
            )
            continue

        if field_max_length and len(translated) > field_max_length:
            logger.warning(
                "Translation exceeds max_length (%d) for %s.%s pk=%s",
                field_max_length, model.__name__, lang_field, obj.pk,
            )
            translated = translated[:field_max_length]

        setattr(obj, lang_field, translated)
        yield lang_field
```
> I haven't tested this.

lang/translation.py
```diff
@@ -200,6 +203,31 @@ def translate_text(self, text, dest_language, source_language=None, table_field=
...
+        source_language = source_language or ""
+        caches = TranslationCache.objects.filter(
+            text_hash=text_hash,
+            source_language=source_language,
+            dest_language__in=dest_languages,
+        )
+        if not caches:
+            return {}
+
+        cache_by_lang = {cache.dest_language: cache for cache in caches}
```
thenav56: `if not caches` will be always be true as caches is a queryset object.

We can instead use cache_by_lang for if


```suggestion
        cache_by_lang = {cache.dest_language: cache for cache in caches}
        if not cache_by_lang:
            return {}
```

thenav56 (changes requested):
Thanks @szabozoltan69 for integrating with cache table.

lang/tasks.py
szabozoltan69: I did it with some more slight changes.

susilnem (approved):

PR #2737, opened 2026-05-18
chore(etl): update the models to include the event_id.

sandeshit:
- Update the models with event_id and parent_event_id.
- Migrations
- Move fetch url to helpers.

## Changes
- Use the new links logic.
- Update the models with event_id and parent_event_id.
- Migrations
- Move fetch url to helpers.

- [X] Updated/ran unit tests
- [X] Updated CHANGELOG.md


alert_system/models.py
```diff
@@ -156,12 +145,12 @@ class ExtractionItem(BaseItem):
     Model for Extraction items.
     """
 
-    class CollectionType(models.IntegerChoices):
-        EVENT = 100, _("event")
-        HAZARD = 200, _("Hazard")
-        IMPACT = 300, _("Impacts")
+    class CollectionType(models.TextChoices):
+        EVENT = "event", _("event")
```
ranjan-stha: _("Event")

alert_system/etl/base/extraction.py
```diff
@@ -129,134 +105,158 @@ def _save_stac_item(self, stac_id: str, defaults: Dict) -> Optional[ExtractionIt
...
-        for feature in impact_features:
-            impact_id = feature.get("id", None)
+        for impact_url in impact_urls:
+            try:
+                impact_item = fetch_stac_data(
+                    impact_url,
+                )
+            except Exception as e:
+                logger.warning(f"Failed to fetch impacts for event {stac_obj.stac_id}: {e}")
+                return []
```
ranjan-stha, line 128: what if only one of the impact_urls fail (let's say 4 out of 5), do we return [ ] or process the success ones and leave behind the failed ones?

alert_system/etl/base/extraction.py
```diff
@@ -272,10 +272,15 @@ def _impact_filter(self, impact_metadata: list[dict]) -> str:
                 and data.get("type") == ImpactDetailsEnum.Type.AFFECTED_TOTAL  # TODO: Add other possible types here.
                 and data.get("value") is not None
             ):
+                value = data["value"]
+                lower_bound = 10 ** (math.floor(math.log10(value)) - 1)
+                upper_bound = 10 ** (math.floor(math.log10(value)) + 1)
+
```
ranjan-stha: what if the value is 0, you have only check for data.get("value") being not None, but not 0, the log(0) ~ undefined. Is this lower_bound/upper_bound justified?

alert_system/etl/gdacs_cyclone/transform.py
```diff
@@ -29,23 +29,42 @@ def compute_buildings_exposed(self, metadata_list) -> Optional[int]:
...
     def process_impact(self, impact_items) -> BaseTransformerClass.ImpactType:
+        latest_episode = -1
         metadata = []
+
         for item in impact_items:
             properties = item.resp_data.get("properties", {})
             impact_detail = properties.get("monty:impact_detail", {})
+
+            if properties.get("forecasted"):
+                continue
```
ranjan-stha, line 40: are we not considering the "forecasted" data point ?

ranjan-stha (commented):
LGTM. Please check few comments.

alert_system/etl/gdacs_cyclone/transform.py
sandeshit: We take the latest episode with forecasted = False for impact point

alert_system/etl/base/extraction.py
sandeshit: Current logic is to fail if anyone fails

alert_system/etl/base/extraction.py
sandeshit: The value should ideally never zero. Adding a check just in case

ranjan-stha (approved):

PR #2739, opened 2026-05-18
Optimize DREF3 pagination by appeal_code

szabozoltan69:
This change shifts DREF3 list pagination to operate on appeal_code values before retrieval, so only the requested page of codes is fetched and serialized. 

The new _order_codes() helper defaults to sorting by appeal_code, and supports order_by=created_at or order_by=-created_at based on the first DREF application created_at per code. The _paginate_codes() helper applies offset and limit to that ordered code list rather than to the final row list. As a result, page sizes in rows can vary because one code can expand to multiple stage rows. Stage filtering still happens after retrieval, so it can further reduce the number of rows returned in a page. 

The OpenAPI schema for the list endpoint now documents the order_by query parameter and its accepted values. 

Overall, the endpoint avoids building the full dataset just to slice it, reducing work and memory for typical paged requests.

### DIFF
--- a/dref/views.py
+++ b/dref/views.py
@@ -8,7 +8,7 @@ from django.db import models, transaction
 from django.http import HttpResponse
 from django.templatetags.static import static
 from django.utils.translation import gettext
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import (
     mixins,
     permissions,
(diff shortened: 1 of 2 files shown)


szabozoltan69:
We switched from sorted(combined) to list(combined) because ordering is now handled later by _order_codes(). That helper applies the default appeal_code sort or the optional order_by=created_at logic, so sorting at the combine step would be redundant and could conflict with order_by.

dref/views.py
```diff
@@ -592,6 +638,9 @@ def list(self, request):
             if excluded_codes:
                 codes = [c for c in codes if c and c.upper() not in excluded_codes]
 
+        codes = self._order_codes(codes, request)
+        codes = self._paginate_codes(codes, request)
```
thenav56, line 647: Hey @szabozoltan69,

Is this the expected?

We're now applying pagination based on the appeal codes, rather than on the items returned in the response. This seems different from the behavior of the standard API, where pagination are applied to the response dataset itself.

If that's correct, then:

- `limit=5` would not necessarily return 5 response items, since the number of items returned for each appeal code can vary.
- `offset=5` would effectively skip the first 5 appeal codes and return results generated from `appeals[5:]`, rather than applying the offset to the final response dataset (`response[5:]`).

szabozoltan69: Hey @thenav56,

Yes, the consequences are exactly what you wrote.

In my view, the performance gains achieved by minimizing resource consumption during pagination far outweigh the disadvantage of having a variable number of records per page (which depends solely on the appeal_codes).

I believe the primary purpose of pagination is to optimize resource utilization and chunk data; users will ultimately iterate through the resulting records anyway, rather than processing a strictly fixed number of rows. Therefore, the benefits of this approach significantly exceed the drawbacks when compared to traditional API conventions.

thenav56: Sounds good.

It should be fine if we document this in the codebase.

szabozoltan69: Thanks @thenav56 
What part of the code (or docs subdir?) should it be documented?

thenav56: Something like this should do
```suggestion
        # NOTE: Ordering and pagination are applied to the appeal codes, not the response objects.
        # As a result, the number of response items may vary. This is expected behavior.
        # This is a temporary limitation. In the future, we plan to apply standard
        # ordering and pagination to the response objects table once an optimized
        # layer is available to support it.
        codes = self._order_codes(codes, request)
        codes = self._paginate_codes(codes, request)
```

szabozoltan69:
Deployed to Staging.
Thanks for the review @thenav56

PR #2754, opened 2026-06-01
feat: Trigger event translation for approved DREF

susilnem:
## Changes
- trigger translation for the emergency generated from DREF
- Add region on event based on the dref country

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release

### DIFF
--- a/dref/utils.py
+++ b/dref/utils.py
@@ -67,7 +67,12 @@ def create_event_from_dref(dref: Dref) -> Event:
         source=Event.EventSource.DREF,
     )
 
-    event.countries.add(dref.country)
+    country = getattr(dref, "country", None)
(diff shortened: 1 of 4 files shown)


dref/utils.py
```diff
@@ -69,4 +69,5 @@ def create_event_from_dref(dref: Dref) -> Event:
 
     event.countries.add(dref.country)
     event.districts.add(*dref.district.all())
+    event.regions.add(dref.country.region)
```
sudip-khanal: There are some countries which are not associated with any region, we might need to add a check for whether the region exists or not.

sudip-khanal (approved):

PR #2759, opened 2026-06-04
Fix Active Dref Operations List n+1 query issue (active-dref)

szabozoltan69:
Refers to Sentry issue /organizations/ifrc-go/issues/6882.

Maybe these changes makes the code too unreadable, and the benefit is not so much. Should we go on with merge?

### DIFF
--- a/dref/serializers.py
+++ b/dref/serializers.py
@@ -252,24 +252,41 @@ class MiniDrefSerializer(serializers.ModelSerializer):
 
     @extend_schema_field(MiniOperationalUpdateActiveSerializer(many=True))
     def get_operational_update_details(self, obj):
-        queryset = DrefOperationalUpdate.objects.filter(dref_id=obj.id).order_by("-created_at")
+        prefetched = getattr(obj, "prefetched_operational_updates", None)
+        queryset = prefetched if prefetched is not None else obj.drefoperationalupdate_set.order_by("-created_at")
         return MiniOperationalUpdateActiveSerializer(queryset, many=True).data
 
     @extend_schema_field(MiniDrefFinalReportActiveSerializer)
     def get_final_report_details(self, obj):
-        queryset = DrefFinalReport.objects.filter(dref_id=obj.id).first()
-        return MiniDrefFinalReportActiveSerializer(queryset).data
+        prefetched = getattr(obj, "prefetched_final_report", None)
+        if prefetched is not None:
+            if isinstance(prefetched, list):
+                final_report = prefetched[0] if prefetched else None
+            else:
+                final_report = prefetched
+        else:
+            try:
+                final_report = obj.dreffinalreport
+            except DrefFinalReport.DoesNotExist:
+                final_report = None
+        if final_report is None:
+            return None
+        return MiniDrefFinalReportActiveSerializer(final_report).data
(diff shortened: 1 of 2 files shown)


dref/views.py
```diff
@@ -345,7 +346,23 @@ class ActiveDrefOperationsViewSet(viewsets.ReadOnlyModelViewSet):
...
+        .prefetch_related(
+            "planned_interventions",
+            "needs_identified",
+            "national_society_actions",
+            "users",
+            Prefetch(
+                "dreffinalreport",
+                queryset=DrefFinalReport.objects.all(),
+                to_attr="prefetched_final_report",
+            ),
```
susilnem, line 359: This is 1:1 Field, So no need to Prefetch DrefFinalReport.

dref/views.py
```diff
@@ -345,7 +346,23 @@ class ActiveDrefOperationsViewSet(viewsets.ReadOnlyModelViewSet):
     permission_classes = [permissions.IsAuthenticated, DenyGuestUserPermission]
     filterset_class = ActiveDrefFilterSet
     queryset = (
-        Dref.objects.prefetch_related("planned_interventions", "needs_identified", "national_society_actions", "users")
+        Dref.objects.select_related("country")
```
susilnem, line 349:
```suggestion
        Dref.objects.select_related("country", "dreffinalreport",)
```

dref/serializers.py
```diff
@@ -278,10 +295,28 @@ def get_application_type_display(self, obj) -> str:
         return gettext("DREF application")
 
     def get_unpublished_op_update_count(self, obj) -> int:
+        prefetched = getattr(obj, "prefetched_operational_updates", None)
+        if prefetched is not None:
+            return sum(1 for op in prefetched if op.status != Dref.Status.APPROVED)
         return DrefOperationalUpdate.objects.filter(dref_id=obj.id).exclude(status=Dref.Status.APPROVED).count()
 
     def get_unpublished_final_report_count(self, obj) -> int:
```
susilnem, line 303: Do we need this field? Given that DrefFinalReport is a OneToOneField, the count can only ever be 0 or 1, so it doesn't seem to provide much additional information.

 If it's required for backward compatibility, we could annotate it in the queryset instead of calculating it in the serializer.

dref/serializers.py
```diff
@@ -252,24 +252,41 @@ class Meta:
...
+        else:
+            try:
+                final_report = obj.dreffinalreport
+            except DrefFinalReport.DoesNotExist:
+                final_report = None
+        if final_report is None:
+            return None
+        return MiniDrefFinalReportActiveSerializer(final_report).data
 
     def get_has_ops_update(self, obj) -> bool:
```
susilnem, line 276: We can shift most of these checks on the query level.

susilnem (changes requested):

szabozoltan69:
We can drop this fix.
Before the change:
kép
After the change, less queries in longer time!
kép

szabozoltan69:
Not effective n+1 query fix.
Nevertheless, if you have suggestions, dear @susilnem or @batpad , maybe about adding "dreffinalreport", let me know and let's do another fix.

PR #2760, opened 2026-06-10
Add external source fields in event 

sudip-khanal:
## Changes
- Add `auto_generated_external_source` field to Event
- Add `auto_generated_external_source_id` field to Event
-  Add `external_source` field to FieldReport
- Add `external_source_id` field to FieldReport
- Set both fields during automatic event creation from a FieldReport

## Checklist

- [X] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


api/models.py
```diff
@@ -828,7 +828,22 @@ class Event(models.Model):
...
+    # External system or source from which the event data originated
+    auto_generated_external_source = models.CharField(
+        verbose_name=_("external source"),
+        max_length=100,
+        null=True,
+        blank=True,
+        help_text=_("External source or system from which the original event data originated."),
+    )
+    # Identifier of the originating event record in the external system
+    auto_generated_external_source_id = models.CharField(
```
susilnem, line 840: Any reason for using CharField here? 

This looks like it should be a UUID field. 

If this value is always expected to be a UUID, would it be possible to use a UUIDField instead? That would provide built-in validation.

api/models.py
```diff
@@ -828,7 +828,22 @@ class Event(models.Model):
     auto_generated_source = models.CharField(
         verbose_name=_("auto generated source"), max_length=50, null=True, blank=True, editable=False
     )
-
+    # External system or source from which the event data originated
+    auto_generated_external_source = models.CharField(
```
susilnem, line 832: Do we even need these fields here in the event table? Do we need filtering for this features?

susilnem (changes requested):
Just some queries!

api/models.py
sudip-khanal: In the MRCS site, we are not using UUIDs for the id field. If it’s coming externally as a UUID, we can switch to a UUIDField. @frozenhelium  can you also confirm this?

api/models.py
sudip-khanal: Yes, we may need these for filtering in the future. @frozenhelium , can you also confirm this?

frozenhelium: That's correct!

api/models.py
frozenhelium: Correct!

PR #2761, opened 2026-06-10
Settings: enable the system to run in submodule

frozenhelium:
- Make git metadata reads optional at startup
- Make compose .env optional and overridable


thenav56 (approved):
LGTM

Let's add a TODO to add a django check to show warning for fallback values

sudip-khanal (approved):

PR #2763, opened 2026-06-12
Fix per-assessment n+1 query issue

szabozoltan69:
Refers to Sentry /organizations/ifrc-go/issues/6969

Dropping from 1169 queries to 11 at around 49ms (instead of 942ms) – the N+1 path is fully addressed.


thenav56 (approved):
Thanks Zoltan.

Changes LGTM

susilnem (approved):

PR #2771, opened 2026-06-29
Dref-Summary: Summary generation task

sandeshit:
## Changes

* Summary generation task
* Retrigger from admin panel

## Checklist

- [X] Updated/ran unit tests
- [X] Updated CHANGELOG.md


dref/tasks.py
```diff
@@ -57,6 +59,72 @@ def send_dref_email(dref_id, users_emails, new_or_updated=""):
...
+    source_id = source_doc.id
+
+    generator = DrefSummaryGenerator()
+    source_hash = generator.compute_source_hash(source_doc)
+
+    summary_instance = DrefSummary.objects.filter(dref=dref).first()
+    if (
+        summary_instance
+        and not overwrite
+        and summary_instance.hash == source_hash
```
susilnem: field mismatch here!

```suggestion
        and summary_instance.source_hash == source_hash
```

Also, rename the field to source_hash instead on prompt_hash on model. That name makes much more sense, looks like we're caching the source data rather than the overall prompt.

dref/models.py
```diff
@@ -1719,6 +1719,21 @@ class SummaryStatus(models.IntegerChoices):
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
+    # The source document the summary was last generated from. Stored so a
+    # retrigger (e.g. from the admin) can replay the same source instead of
+    # re-resolving the latest approved document.
+    source_model_name = models.CharField(
```
susilnem: Better to create a enum for this instead of saving on charfield?

dref/tasks.py
```diff
@@ -57,6 +59,72 @@ def send_dref_email(dref_id, users_emails, new_or_updated=""):
...
+    """Generate and store the AI-assisted summaries for a DREF."""
+    dref = Dref.objects.filter(id=dref_id).first()
+    if not dref:
+        logger.error(f"DREF not found for summary generation: ({dref_id})")
+        return False
+
+    if source_model_name and source_id:
+        try:
+            source_model = apps.get_model(source_model_name)
+            source_doc = source_model.objects.filter(id=source_id).first()
```
susilnem:
```suggestion
            source_instance = source_model.objects.filter(id=source_id).first()
```

dref/summary.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+class DrefSummaryGenerator:
+    """Assembles per-section prompts from a source document and produces all summaries."""
+
+    def __init__(self, client: Optional[DrefSummaryLLMClient] = None):
+        self.client = client or DrefSummaryLLMClient()
+
+    @classmethod
+    def _model_label(cls, source_doc) -> str:
```
susilnem: There's already a utils function forthis! Use that instead.

dref/tasks.py
```diff
@@ -57,6 +59,72 @@ def send_dref_email(dref_id, users_emails, new_or_updated=""):
...
+        except Exception:
+            logger.error(
+                f"Could not resolve source model ({source_model_name}) for DREF ({dref_id}) summary",
+                exc_info=True,
+            )
```
susilnem: For the logger error, Use logger_context for dynamic data, it helps with grouping in Sentry!

Something like this:

```
logger.error(
    "Could not resolve source model for DREF summary",
    extra=logger_context({"source_model_name": source_model_name, "dref_id": dref_id}),
    exc_info=True,
)
```

dref/tasks.py
```diff
@@ -57,6 +59,72 @@ def send_dref_email(dref_id, users_emails, new_or_updated=""):
 ]
 
 
+@shared_task
+def generate_dref_summary(dref_id, source_model_name=None, source_id=None, overwrite=False):
```
susilnem: Do we need `dref_id` as a separate param here? 
We already get source_model_name + source_id, we could derive the dref from that instead:

  - If source is Dref, dref_id = source_id.
  - If source is DrefOperationalUpdate or DrefFinalReport, both have a dref field, so we can just use source_doc.dref_id.

  This avoids passing two things that should always match but currently aren't checked against each other.

dref/models.py
susilnem, line 1: Add status, source any other relavant fields on the serializer!

dref/tasks.py
```diff
@@ -57,6 +59,72 @@ def send_dref_email(dref_id, users_emails, new_or_updated=""):
...
+        and not overwrite
+        and summary_instance.hash == source_hash
+        and summary_instance.status == DrefSummary.SummaryStatus.SUCCESS
+    ):
+        logger.info(f"DREF summary up to date for DREF ({dref_id}); skipping generation.")
+        return True
+
+    if summary_instance is None:
+        summary_instance = DrefSummary(dref=dref)
+    summary_instance.hash = source_hash
```
susilnem: Here too!

susilnem (changes requested):
Overall looks good, just some minor changes.

It's more better if we can add test cases for this flow?

dref/models.py
```diff
@@ -1710,6 +1710,11 @@ class SummaryStatus(models.IntegerChoices):
         SUCCESS = 300, _("Success")
         FAILED = 400, _("Failed")
 
+    class SourceModel(models.TextChoices):
```
susilnem: Better to use Integer field!

dref/models.py
```diff
@@ -1719,7 +1724,23 @@ class SummaryStatus(models.IntegerChoices):
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
-    prompt_hash = models.CharField(
+    # The source document the summary was last generated from. Stored so a
+    # retrigger (e.g. from the admin) can replay the same source instead of
+    # re-resolving the latest approved document.
+    source_model_name = models.CharField(
```
susilnem: This one should not be null.

dref/models.py
```diff
@@ -1719,7 +1724,23 @@ class SummaryStatus(models.IntegerChoices):
...
+        max_length=255,
+        verbose_name=_("source model name"),
+        choices=SourceModel.choices,
+        blank=True,
+        null=True,
+    )
+    source_id = models.PositiveBigIntegerField(
```
susilnem, line 1738: This one too!

susilnem (changes requested):

dref/summary.py
```diff
@@ -0,0 +1,388 @@
...
+
+        results = dict(empty_results)
+        for field_name in SUMMARY_FIELDS:
+            value = parsed.get(field_name)
+            if not isinstance(value, str):
+                continue
+            summary = value.strip()
+            if len(summary) > MAX_OUTPUT_CHARS_PER_FIELD:
+                summary = summary[:MAX_OUTPUT_CHARS_PER_FIELD].rstrip()
+            results[field_name] = summary
```
susilnem, line 387: I dont think, Translation is triggered here! can you verify this?

susilnem (approved):

PR #2773, opened 2026-07-01
Alert-system: Map to internal cluster eoapi url

sandeshit:
## Changes

* Map external urls for related events to internal cluster eoapi url.
* Avoid raise on empty data.

## Things to remember

We need to set two env for this change.
for these values, take reference from montandon configurations
```
EOAPI_STAC_INTERNAL_URL=
EOAPI_STAC_EXTERNAL_URL=
```

## Checklist

- [x] Updated/ran unit tests
- [x] Updated CHANGELOG.md

## NOTE
- This is a breaking change in staging environment.

### DIFF
--- a/alert_system/helpers.py
+++ b/alert_system/helpers.py
@@ -10,8 +10,8 @@ def _remap_stac_url(url: str) -> str:
     """
     Rewrite a STAC URL to point at the internal cluster service.
     """
-    external_base = getattr(settings, "EOAPI_STAC_EXTERNAL_URL", None)
-    internal_base = getattr(settings, "EOAPI_STAC_INTERNAL_URL", None)
+    external_base = settings.EOAPI_STAC_EXTERNAL_URL
+    internal_base = settings.EOAPI_STAC_INTERNAL_URL
     if not external_base or not internal_base:
         return url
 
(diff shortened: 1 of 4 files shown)


alert_system/helpers.py
```diff
@@ -1,7 +1,32 @@
...
+    ext = urlsplit(external_base.rstrip("/"))
+    internal = urlsplit(internal_base.rstrip("/"))
+    parsed = urlsplit(url)
+
+    if parsed.netloc not in (ext.netloc, internal.netloc):
+        return url
+
+    path = parsed.path
+    if ext.path and path.startswith(ext.path):
+        path = path[len(ext.path) :]
```
susilnem: May be something like this?


```suggestion
    if ext.path and (path == ext.path or path.startswith(ext.path + "/")):
        path = path[len(ext.path):]
```

susilnem (changes requested):
@sandeshit Can you verify this changes?

susilnem (approved):

susilnem:
Deployed to staging and update values from montandon configs.

https://github.com/IFRCGo/go-deploy/blob/754e9a445e5a891f156bd4174ee60fa102c37b1e/applications/go-api/helm/ifrcgo-helm/values-staging.yaml#L27-L28

PR #2784, opened 2026-07-11
fix(lang): skip failed translation fields on API errors (#2303)

### LINKED ISSUE #2303: Handle translation server error
## Issue
If there is 500 error response from the translation server side, go-api skip the translation of the whole object instead of just that field/request. We should properly handle this be skipping the trouble field/request.

Related sentry error: https://sentry.northeurope.cloudapp.azure.com/organizations/ifrc-go/issues/2066/?project=2&referrer=issue-stream&statsPeriod=30d&stream_index=5

Tarekchehahde:
## Summary
- On IFRC translation HTTP errors (including 500), log `text_length` and return `None` instead of attempting JSON parse
- Wrap Amazon Translate calls similarly so a single field failure does not abort other languages
- `ModelTranslator` already skips `None` translations per field

## Test plan
- [x] Unit test: IFRC 500 response returns `None`
- [ ] Verify Sentry logs include `text_length` extra on production errors

Fixes #2303

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lang/tests.py
+++ b/lang/tests.py
@@ -266,6 +266,23 @@ class TranslatorMockTest(unittest.TestCase):
             with override_settings(TESTING=False):
                 assert ifrc_translator.translate_text("hello", "es") == "Hola"
 
+    @pytest.mark.django_db
+    @mock.patch("lang.translation.requests")
+    def test_ifrc_translator_server_error_returns_none(self, requests_mock):
+        response_mock = mock.Mock()
+        response_mock.status_code = 500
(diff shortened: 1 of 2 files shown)


Tarekchehahde:
Follow-up: [#2790](https://github.com/IFRCGo/go-api/pull/2790) adds recursive HTML chunking for ~50k+ fields (same #2303). Happy to close #2784 in favor of #2790 or keep both if you prefer the smaller error-handling-only change first.

PR #2813, opened 2026-08-17
Feat: Add celery beat for scheduling new cronjobs

susilnem:
## Description
Introduce django-celery-beat so new cronjobs can be declared in code and
dispatched by a beat process, instead of adding another Kubernetes CronJob.

Existing cronjobs are untouched. The 30 legacy jobs keep running as k8s
CronJobs from values.yaml:cronjobs and stay monitored via SentryMonitor +
`manage.py cron_job_monitor`. A job belongs to one mechanism or the other,
never both -- SentryMonitor.validate_config() asserts the enum matches
values.yaml.

## Changes
- Introduce django-celery-beat
- main/cronjobs.py: SCHEDULES registry, kombu queues (default/heavy/cronjob) feeding task_queues, per-job Sentry monitor config, and a beat_init hook that prunes PeriodicTask rows for removed jobs ("manual:" rows spared)
- seed jobs: clear_expired_django_sessions (django_session was never pruned) and a per-queue celery_queue_uptime_check heartbeat
- main/sentry.py: SentryConfig dataclass, every option passed to sentry_sdk.init() by name so an unrecognised key cannot reach it
- SENTRY_DEBUG decouples Sentry verbosity from DJANGO_DEBUG
- misc/dev: worker and beat entrypoints with a broker wait
- tests in main/test_cronjobs.py, a django check for SCHEDULES task paths, and docs/cronjobs.md

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


api/tasks.py
```diff
@@ -59,3 +61,20 @@ def generate_export_pdf(export_id, title, set_user_language="en"):
         export.status = Export.ExportStatus.ERRORED
         export.save(update_fields=["status"])
     logger.info(f"End export: {export.pk}")
+
+
+# TODO(susilnem): Do we need this cron?
```
thenav56: Yes, https://docs.djangoproject.com/en/6.0/topics/http/sessions/#clearing-the-session-store

docker-compose.yml
```diff
@@ -143,7 +143,19 @@ services:
   # For development only
   celery:
     <<: *base_server_setup
-    command: python manage.py run_celery_dev
+    restart: unless-stopped
+    command: ./misc/dev/run_worker.sh
+    healthcheck:
```
thenav56, line 150: Let's add TODO here for healthcheck to upate after banjo-utils integration

docker-compose.yml
```diff
@@ -143,7 +143,19 @@ services:
...
+      test: ["CMD-SHELL", "celery -A main inspect ping -d celery@$$HOSTNAME || exit 1"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 40s
+
+  celery-beat:
+    <<: *base_server_setup
+    restart: unless-stopped
+    command: ./misc/dev/run_worker_beat.sh
```
thenav56, line 160: Let's add TODO here for healthcheck to upate after banjo-utils integration

thenav56 (approved):
Looks good

api/tasks.py
susilnem: But it would be effective only for the admin panel users. right?

thenav56: Yes sir

PR #2821, opened 2026-09-02
refactor(per): consolidate PER dashboard data endpoints

arunissun:
## Summary

- Consolidate the PER map and dashboard data query/serialization logic.
- Return stable latest-country map results while retaining complete PER process history.
- Add read-only response caching and bounded-query coverage for both endpoints.
- Add focused regression tests for response behavior, ordering, cache use, and realistic related data.

## Validation

- Black, isort, and Flake8 passed.
- Django migration check passed with no changes detected.
- Focused PER endpoint tests: 9 passed.
- OpenAPI schema matches the committed artifact.
- Full API suite: 324 passed, 4 skipped.

## Scope

No migrations, route changes, frontend changes, or unrelated cleanup.

### DIFF
--- a/per/dashboard_data.py
+++ b/per/dashboard_data.py
@@ -3,6 +3,7 @@ from typing import Any
 
 from django.db.models import Prefetch
 
+from .dashboard_utils import AREA_NAMES, contains_affirmative
 from .models import (
     AreaResponse,
     FormComponentResponse,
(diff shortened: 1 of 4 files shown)


szabozoltan69:
Hi @arunissun – it seems to be some redundancy in this change.
E.g. the AREA_NAMES and the AFFIRMATIVE_WORDS are also in per/drf_views.py, and you re-define them in per/dashboard_data.py.
Could you please refactor your code so that such duplications should not be there?

arunissun:
@szabozoltan69 I WILL do that thanks for the review

szabozoltan69:
Deployed to Staging

PR #2824, opened 2026-09-07
Hotfix(emergency): replace joined Count() aggregates with per-relation subqueries

susilnem:
## Problem

`GET /api/v2/emergency/{id}/` was timing out (504) for events with a lot of related data. Root cause: `EmergencyViewset.get_queryset()` computed 3 counts (projects, surge alerts, active deployments) using `Count(distinct=True)` on 3 separate relations inside one query. Joining unrelated one-to-many relations multiplies rows instead of adding them — for event 5854 that's 98 × 302 × 330 = **9.7 million rows** generated just to get 3 numbers, forcing Postgres to sort/group all of it (spilled 6.9 GiB to disk).

## Fix

Replaced the 3 joined `Count()` aggregates with 3 independent correlated subqueries (one per count, no cross-joining). Same result, no row explosion.

## Before / After (event 5854)

| | Before | After |
|---|---|---|
| Response time | timeout (>120s, prod hit 74s → 504) | **0.47s** |
| response_activity_count | (never returned) | 98 |
| active_deployments_count | (never returned) | 0 |
| surge_alerts_count | (never returned) | 302 |

## Before / After (event 6646, smaller: old code could finish here)

| | Before | After |
|---|---|---|
| Response time | 29.65s | **0.28s** |
| response_activity_count | 7 | 7 |
| active_deployments_count | 0 | 0 |
| surge_alerts_count | 72 | 72 |

Values identical before/after:  confirms this is a performance fix only, no behavior change.

### DIFF
--- a/api/drf_views.py
+++ b/api/drf_views.py
@@ -59,13 +59,14 @@ from api.visibility_class import (
 )
 from country_plan.models import CountryPlan
 from databank.serializers import CountryOverviewSerializer
-from deployments.models import ERU, Personnel
+from deployments.models import ERU, EmergencyProject, Personnel
 from deployments.serializers import ListDeployedERUByEventSerializer
 from dref.models import Dref, DrefFinalReport, DrefOperationalUpdate
 from main.enums import GlobalEnumSerializer, get_enum_values
 from main.filters import NullsLastOrderingFilter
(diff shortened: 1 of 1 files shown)


szabozoltan69 (approved):

szabozoltan69:
Quick and great solution!

PR #2826, opened 2026-09-11
fix(docker): build on debian bookworm

thenav56:
## Addresses
- Bullseys's EOL - https://www.debian.org/News/2026/20260831


thenav56:
https://github.com/IFRCGo/go-api/pull/2827

PR #2829, opened 2026-09-15
fix(ingest_appeals): handle null sector/country from upstream API

susilnem:
## Changes

- sector: keep existing value when OSS_name is null instead of overwriting it
- country: split multi-country codes and resolve by the last one; fall back to the IFRC placeholder if still unresolved

## Checklist

- [ ] Updated/ran unit tests
- [ ] Updated CHANGELOG.md

## Release


susilnem:
@szabozoltan69 , please update the changes if needed.

szabozoltan69 (approved):
Beautiful and efficient!

PR #2831, opened 2026-09-20
chore(migrations): squash all app migrations with replaces

thenav56:
## Changes

Squashes all app migrations with `replaces`: **745 files → 14** across 13 apps.
Old migration files are kept, so deployed databases transition automatically —
no faking, no manual `django_migrations` surgery.

`deployments` is squashed in two parts (`0001_squashed_0081`, `0082_squashed_0097`)
because a single squash forms a dependency cycle with `notifications`
(`deployments.0082 → notifications.0013`, `notifications.0008 → deployments.0033`).

## Review checklist

Most of the diff is generated. These parts are hand-edited and deserve eyes:

- [x] **Stale defaults stripped** (11 fields, 5 apps). Django folds
      `AlterField(..., preserve_default=False)` into `CreateModel` but keeps the
      throwaway default. Removed in `local_units` (6), `per` (2), `lang`,
      `deployments`, `dref`. Verify each matches the model.
- [x] **`triggers_to_db` callables inlined** in `api/0001_squashed_0232.py`
      without their raw `BEGIN;`/`COMMIT;`. Those statements desync psycopg2 from
      the server transaction; harmless when each migration had its own
      transaction, but squashed they break `0232`'s `.iterator()` with
      `DECLARE CURSOR can only be used in transaction blocks`.
- [x] **Three `elidable=True` seeds re-added** (`delegation_office_types.json`,
      `health-affiliation-functionality.json`, `organizationtypes.json`).
      `squashmigrations` drops elidable ops by design, which silently removed
      reference data that fresh installs need.
- [x] Data-migration functions inlined from the replaced files are unchanged
      apart from name prefixing.

## Verification

- [x] `python manage.py makemigrations --check --dry-run` → `No changes detected`
- [x] Migration graph loads (no `CircularDependencyError`)
- [x] CI green

**Against a restored staging/production copy:**

- [x] `python manage.py migrate --plan` → `No planned migration operations.`
- [ ] `python manage.py migrate` → `No migrations to apply.`
- [x] All 14 squashed migrations now recorded (Django's `check_replacements()`
      inserts these during `migrate`; the cleanup release depends on them):

```sql
select app, name from django_migrations
where app in ('alert_system','api','country_plan','databank','deployments','dref',
              'eap','flash_update','lang','local_units','notifications','per',
              'registrations')
  and name ~ '^(0001|0082)_squashed_'
order by app, name;   -- expect exactly 14 rows
```

**From an empty database:**

- [x] `python manage.py migrate` succeeds
- [x] Seed tables populated: `local_units_affiliation` (6),
      `local_units_delegationofficetype` (7), `per_organizationtypes` (2)
- [x] Schema equivalent to a DB migrated from the unsquashed tree

## Known, accepted difference

On **new** databases only, 70 constraints and 72 indexes get different *names* —
definitions are identical and there are zero column differences. Two causes:
unique constraints folded into `CreateModel` take Postgres's default `_key` name
instead of Django's `_uniq`; and constraints on later-renamed models hash from
the final name (e.g. the `eap_eaprefrence` → `eap_eapreference` typo fix).
Django resolves constraints by introspection, not by name. Existing databases are
unaffected — they never re-run the squash.

## Deploy order — required

- [x] Deploy this to **every** environment and confirm the 14 rows above
- [x] Only then merge/deploy the follow-up cleanup PR (deletes the 745 replaced
      files and removes `replaces`)

Applying the cleanup release to a database that has not applied this one makes
Django run the squashed migrations from scratch against a populated database.
Reproduced: it fails with `relation "api_action" already exists`.


szabozoltan69:
I desire this line sooo much for a long time! Many thanks, dear Navin!!! ✨️

szabozoltan69 (approved):
Great job!

PR #2832, opened 2026-09-22
chore(migrations): drop migrations replaced by the squash

thenav56:
Deletes the 745 replaced files, removes the `replaces` lists, and repoints cross-app dependencies at the squashed migrations.


szabozoltan69 (approved):
Formidable!

